PR #8, opened 2020-04-28
Add example from emulator

Kay-Zee:
This takes the `fungible_token_test.go` from the emulator repo. The tests don't currently pass since the FT contract from that repo is different than this one. Will need to change up the tests a bit.


joshuahannan:
@Kay-Zee Looks like there the travis build is failing because of an authentication error with the dapperlabs repo. Any idea of how to fix that?

PR #13, opened 2020-05-19
Add contracts module containing FungibleToken/FlowToken

benjaminkvm:
## Description

We currently use contract/contract interfaces from this repository in other repositories for testing, however we have them copied versus imported from a single source of truth (this repository).

This PR adds a `contracts` go module to this repository, containing two methods:

- `GenerateFungibleTokenContractInterface` which returns the `FungibleToken` contract interface.
- `GenerateFlowTokenContract` which returns the `FlowToken` contract.

We should probably also do this with the onflow/flow-nft repository.

Notes: 
- I created a new module versus using the existing `tests` module (wasn't sure other repositories would need the methods that currently exist there), let me know if it would be preferable to just re-use that module.
- I see the `go.mod` has the go version set at 1.14, I can set it to 1.13 if we are still on that version.


joshuahannan:
Is there a way we could have it generate directly from the files in `contracts/`? The way you have it in the PR still requires us to copy/paste the contracts to the generate functions every time we make changes.

benjaminkvm:
Yeah I was thinking about that!

I removed the `Dummy` contract from the method that returns the `FungibleToken` contract interface, so it just returns the contract interface and it is up to the caller to implement the contract if they choose (just wanted to give some additional flexibility there). I could add it back and just slurp the file.

Also, I templated out the address of the `FungibleToken` contract interface in the `FlowToken` contract, could find/replace on slurp or just leave it hardcoded.

joshuahannan:
We could just do it like Kan did it here:

https://github.com/onflow/flow-go-sdk/blob/kan/ft-example-test-net/examples/fungible_token/main.go#L42-L47
and then
https://github.com/onflow/flow-go-sdk/blob/kan/ft-example-test-net/examples/fungible_token/main.go#L463-L474

and just template out all the things that make the FlowToken unique, like the import address, the FlowToken names, and such

benjaminkvm:
Updated this PR with recommendations made earlier! I noticed it failed CI, however it appears `master` has also?

joshuahannan:
yeah @benjaminkvm CI is still broken on these repos unfortunately

PR #16, opened 2020-05-23
added FlowToken.Vault type to all capability restrictions

joshuahannan:
We realized that in order for users to have a guarantee that the Vault they are depositing to or reading the balance from is a FlowToken Vault, we need to restrict the types of the Capabilities to be `FlowToken.Vault{FungibleToken.Receiver}` instead of just `AnyResource{FungibleToken.Receiver}`


joshuahannan:
@Kay-Zee This includes an update to the init function of the FlowToken contract. I just wanted to make you aware of the change in case it needed to be updated anywhere else for testing

PR #17, opened 2020-05-23
Bundle contract code using bindata tool

psiemens:
- Use `github.com/kevinburke/go-bindata` to embed contract code into the `flow-ft/contracts` Go module
- Add Makefile with `test`, `generate` and `check-generated` targets
- Add CI workflow that uses `check-generated` to make sure embedded code matches the source
- Remove `.travis.yml`, replace with GitHub CI
  - Temporarily disable main unit tests until `flow-emulator` is public
- Consolidate all Cadence code into `src` (I can undo this, just thought it was a bit cleaner)


contracts/contracts.go
```diff
@@ -1,59 +1,37 @@
...
-	fungibleTokenCode []byte
-	flowTokenTemplate []byte
-)
-
-// Load the contracts when the module is loaded.
-func init() {
-	fungibleTokenCode, _ = getContract(FungibleTokenContractFilename)
-	flowTokenTemplate, _ = getContract(FlowTokenContractFilename)
+// FungibleToken returns the FungibleToken contract interface.
+func FungibleToken() ([]byte, error) {
```
psiemens, line 18: I removed the `Contract` suffix on these functions because it seemed a bit redundant given the package name, but I can add it back!

```
import "github.com/onflow/flow-ft/contracts`

...

ft := contracts.FungibleToken()

joshuahannan: This is fine with me. 👍

PR #19, opened 2020-05-24
Introduce FlowToken.Administrator resource

psiemens:
**Note: tests will fail on CI and locally because this branch depends on a special version of the emulator.**

Due to the limitation of one contract per account, we need a way to delegate administrator privileges over the `FlowToken` contract to another account. This will allow the special Flow service account to control the Flow token supply.

This PR introduces the `FlowToken.Administrator` resource, which is the only object with the capability to create `FlowToken.Minter` resources. The `FlowToken.Administrator` is placed into the `admin` account upon initialization of the `FlowToken` contract.

This PR also splits `MintAndBurn` into two separate resources, `Minter` and `Burner`. Is there a particular reason they were being combined into the same resource? I mainly made this change while I was trying to better understand the code, so I can easily reverse it.

Lastly, this PR updates `FlowToken.init` to set an initial balance of `0.0`, allowing the `admin` account to choose when and how many tokens to mint.

## Open Questions

- I don't like that `admin: AuthAccount` is passed as an argument to `FlowToken.init`. Is there a better way to do this? 
  - Idea: store `FlowToken.Administrator` in the `FlowToken` storage, and then use a second transaction to move it to another account. This requires us to deploy `FlowToken` with a key, make the transfer, and then revoke the key.


src/contracts/FlowToken.cdc
```diff
@@ -169,38 +170,34 @@ pub contract FlowToken: FungibleToken {
...
-        // Initialize the totalSupply field to the initial balance
-        self.totalSupply = 1000.0
+    init(adminAcct: AuthAccount) {
+        self.totalSupply = 0.0
 
         // Create the Vault with the total supply of tokens and save it in storage
         //
         let vault <- create Vault(balance: self.totalSupply)
-        self.account.save(<-vault, to: /storage/flowTokenVault)
+        adminAcct.save(<-vault, to: /storage/flowTokenVault2)
```
psiemens, line 179: Using `flowTokenVault2` for now to avoid a conflict with the bootstrapped `flowTokenVault` resource that the emulator places in each account.

src/contracts/FlowToken.cdc
```diff
@@ -169,38 +170,34 @@ pub contract FlowToken: FungibleToken {
     //
     // The numbers are arbitrary.
     //
-    init() {
-        // Initialize the totalSupply field to the initial balance
-        self.totalSupply = 1000.0
+    init(adminAcct: AuthAccount) {
+        self.totalSupply = 0.0
```
psiemens, line 174: If we are delegating minting privileges to another account, I don't really think it makes to mint an initial supply into a vault that nobody can access.

src/contracts/FlowToken.cdc
```diff
@@ -134,6 +139,15 @@ pub contract FlowToken: FungibleToken {
...
         }
 
+        init(allowedAmount: UFix64) {
+            self.allowedAmount = allowedAmount
+        }
+    }
+    
+    // Burner
+    //
+    pub resource Burner {
```
psiemens, line 149: Minting and burning felt like separate actions to me, so I decoupled them. I might be missing something, though!

src/contracts/FlowToken.cdc
```diff
@@ -108,12 +94,31 @@ pub contract FlowToken: FungibleToken {
...
+
+        // createNewBurner
+        //
+        // Function that creates and returns a new burner resource
+        //
+        pub fun createNewBurner(): @Burner {
+            emit BurnerCreated()
+            return <-create Burner()
+        }
+    }
```
psiemens, line 115: Only the `Administrator` can create `Minter` and `Burner` resources.

test/token_test.go
```diff
@@ -412,3 +415,105 @@ func TestMintingAndBurning(t *testing.T) {
...
+	// Should be able to deploy a contract as a new account with no keys.
+	flowTokenCode := readFile(flowTokenContractFile)
+
+	setupScript := fmt.Sprintf(`
+      transaction {
+        prepare(signer: AuthAccount) {
+          let flowTokenAcct = AuthAccount(payer: signer)
+
+          let admin = signer
+          flowTokenAcct.setCode("%s".decodeHex(), admin)
```
psiemens, line 439: Ideally we'd be able to do something like this, but unfortunately we can't import `FlowToken` in this transaction because it doesn't exist yet statically.

```
flowTokenAcct.setCode("%s".decodeHex())

let admin <- flowTokenAcct.load<FlowToken.Administrator>(from: /storage/flowTokenAdmin)
signer.save<FlowToken.Administrator>(<-admin, to: /storage/flowTokenAdmin)

File: test/token_test.go [hunk 6]
```diff
@@ -412,3 +415,105 @@ func TestMintingAndBurning(t *testing.T) {
...
+          let minter <- self.tokenAdmin.createNewMinter(allowedAmount: 100.0)
+          let mintedVault <- minter.mintTokens(amount: 100.0)
+
+          self.tokenReceiver.deposit(from: <-mintedVault)
+
+          log("Minted 100 tokens and deposited to admin account")
+
+          destroy minter
+        }
+      }
```
psiemens, line 495: This transaction template is used by the token admin to mint tokens, which it can then transfer to whoever it wants.

src/contracts/FlowToken.cdc
joshuahannan: I think it is fine to split them up. I put them together because I thought that it would be the same accounts doing both, but this is fine

src/contracts/FlowToken.cdc
joshuahannan: This is fine. As long as the correct initial supply of tokens get minted at genesis

test/token_test.go
joshuahannan: yeah I don't think there is any way to do it like this

src/contracts/FlowToken.cdc
```diff
@@ -169,38 +170,34 @@ pub contract FlowToken: FungibleToken {
     //
     // The numbers are arbitrary.
     //
-    init() {
-        // Initialize the totalSupply field to the initial balance
-        self.totalSupply = 1000.0
+    init(adminAcct: AuthAccount) {
```
turbolent, line 173: Nit: Can we maybe spell this out and name this `adminAccount`?

test/token_test.go
```diff
@@ -412,3 +415,105 @@ func TestMintingAndBurning(t *testing.T) {
...
+      import FungibleToken from 0x02
+      import FlowToken from 0x06
+
+      transaction {
+        let tokenAdmin: &FlowToken.Administrator
+        let tokenReceiver: &FlowToken.Vault{FungibleToken.Receiver}
+
+        prepare(signer: AuthAccount) {
+          self.tokenAdmin = signer.
+            borrow<&FlowToken.Administrator>(from: /storage/flowTokenAdmin)
```
turbolent, line 476:
```suggestion
          self.tokenAdmin = signer
            .borrow<&FlowToken.Administrator>(from: /storage/flowTokenAdmin) 
```

PR #20, opened 2020-05-28
update dependencies and use example token instead of flow token

joshuahannan:
In this PR:

* Updated dependencies for emulator, go-sdk, and cadence
* fixed transaction templates to use the correct minting and burning
* separated deployment into its own function
* Make all the transaction templates generic so that any fungible token could use them and just insert their names and storage locations into the templates.

* Made an exact copy of FlowToken as ExampleToken, minus the initializer that requires an argument.  Now that FlowToken is deployed, I think it should be part of the flow-core-contracts repository and this repo can be specifically for the fungible token standard. I left FlowToken in here in case there is something that depends on it right now, but ideally I would like to get it out of here completely and use example token from now on


test/templates.go
```diff
@@ -10,64 +10,64 @@ import (
...
 
-              acct.link<&FlowToken.Vault{FungibleToken.Receiver}>(/public/flowTokenReceiver, target: /storage/flowTokenVault)
-              acct.link<&FlowToken.Vault{FungibleToken.Balance}>(/public/flowTokenBalance, target: /storage/flowTokenVault)
+              acct.link<&%s.Vault{FungibleToken.Receiver}>(/public/%sReceiver, target: /storage/%sVault)
+              acct.link<&%s.Vault{FungibleToken.Balance}>(/public/%sBalance, target: /storage/%sVault)
           }
       }
     `
-	return []byte(fmt.Sprintf(template, fungibleAddr, flowAddr))
+	return []byte(fmt.Sprintf(template, fungibleAddr, tokenName, tokenAddr, tokenName, storageName, tokenName, storageName, storageName, tokenName, storageName, storageName))
```
psiemens, line 29: You can simplify this a bit with indexes to avoid repeating arguments:

```go
func GenerateCreateTokenScript(fungibleAddr, tokenAddr flow.Address, tokenName, storageName string) []byte {
	template := `
	  import FungibleToken from 0x%[1]s 
	  import %[3]s from 0x%[2]s
      transaction {
          prepare(acct: AuthAccount) {
              let vault <- %[3]s.createEmptyVault()
              acct.save(<-vault, to: /storage/%[4]sVault)
              acct.link<&%[3]s.Vault{FungibleToken.Receiver}>(/public/%[4]sReceiver, target: /storage/%[4]sVault)
              acct.link<&%[3]s.Vault{FungibleToken.Balance}>(/public/%[4]sBalance, target: /storage/%[4]sVault)
          }
      }
    `
	return []byte(fmt.Sprintf(template, fungibleAddr, tokenAddr, tokenName, storageName))
}
```

psiemens: Same for all the templates in this file

test/templates.go
```diff
@@ -77,95 +77,94 @@ func GenerateTransferVaultScript(fungibleAddr, flowAddr flow.Address, receiverAd
...
-				self.vault <- mintAndBurn.mintTokens(amount: %d.0)
+			  self.tokenAdmin = signer
+				.borrow<&%s.Administrator>(from: /storage/%sAdmin) 
+				?? panic("Signer is not the token admin")
+	
+			  self.tokenReceiver = getAccount(0x%s)
+				.getCapability(/public/%sReceiver)!
+				.borrow<&%s.Vault{FungibleToken.Receiver}>()
+				?? panic("Unable to borrow receiver reference")
 			}
```
psiemens, line 103: 👍

joshuahannan:
@psiemens I fixed all the templates. Could I get an approval so I can merge? Thanks! 😄

contracts/go.mod
```diff
@@ -3,7 +3,9 @@ module github.com/onflow/flow-ft/contracts
 go 1.14
 
 require (
+	github.com/dapperlabs/flow-emulator v0.4.0 // indirect
```
psiemens, line 6: Is `github.com/dapperlabs/flow-emulator` needed as a dependency here? Unfortunately it's breaking CI because it' still a private repo. 😢 

Maybe try running `go mod tidy` in this repo?

PR #22, opened 2020-06-05
Adds custom example token contract Getters

joshuahannan:
I added getters for the example contract where anyone can replace the `ExampleToken` name with any name of their choice so others can easily deploy their own custom token. Also added a test to ensure that it works.

Primarily so Dapper Utility Coin can use this to deploy without having to copy and paste


test/go.mod
```diff
@@ -9,3 +9,5 @@ require (
 	github.com/onflow/flow-go-sdk v0.4.0
 	github.com/stretchr/testify v1.5.1
 )
+
+replace github.com/onflow/flow-ft/contracts => ../contracts
```
psiemens, line 13: 👍

contracts/contracts.go
```diff
@@ -51,6 +51,27 @@ func ExampleToken(fungibleTokenAddr string) []byte {
 	return []byte(code)
 }
 
+// CustomizableExampleToken returns the ExampleToken contract with a custom Name.
+//
+// The returned contract will import the FungibleToken contract from the specified address.
+func CustomizableExampleToken(fungibleTokenAddr, tokenName string) []byte {
```
psiemens, line 57:
```suggestion
func CustomToken(fungibleTokenAddr, tokenName string) []byte {
```

Maybe just shorten it for simplicity?

contracts/contracts.go
```diff
@@ -51,6 +51,27 @@ func ExampleToken(fungibleTokenAddr string) []byte {
 	return []byte(code)
 }
 
+// CustomizableExampleToken returns the ExampleToken contract with a custom Name.
+//
+// The returned contract will import the FungibleToken contract from the specified address.
```
psiemens, line 56:
```suggestion
// The returned contract will import the FungibleToken interface from the specified address.
```

I could see this causing a little bit of confusion

contracts/contracts.go
```diff
@@ -51,6 +51,27 @@ func ExampleToken(fungibleTokenAddr string) []byte {
 	return []byte(code)
 }
 
+// CustomizableExampleToken returns the ExampleToken contract with a custom Name.
```
psiemens, line 54:
```suggestion
// CustomizableExampleToken returns the ExampleToken contract with a custom name.
```

PR #24, opened 2020-06-06
invalid transfer token and forwarding

joshuahannan:
* Wrote a test to ensure that you cannot deposit a `Vault` from one token into a `Vault` of a different token
* Made some fixes to tokenforwarding and added a test to ensure it works correctly
* removed the storageName parameter from script generation to prevent mistakes
* made it possible to customize balance in CustomToken


src/contracts/TokenForwarding.cdc
```diff
@@ -26,32 +26,43 @@ pub contract TokenForwarding {
...
+            emit ForwardedDeposit(amount: balance, from: self.owner?.address)
         }
 
         // changeRecipient changes the recipient of the forwarder to the provided recipient
         //
-        pub fun changeRecipient(_ newRecipient: &{FungibleToken.Receiver}) {
+        pub fun changeRecipient(_ newRecipient: Capability) {
+            pre {
+                newRecipient.borrow<&{FungibleToken.Receiver}>() != nil: "Could not borrow Receiver reference from the Capability"
+            }
```
turbolent, line 51: What's the reason for adding this? Ensuring the capability can be borrowed in this setter doesn't really ensure it can be borrowed later. Same below

joshuahannan: Its not perfect, but even though it can be changed later potentially, it is good to know that at least it works when you initially set the value

joshuahannan: Does that make sense @turbolent? Are you ok with me merging it?

turbolent: Yep, that makes sense

PR #29, opened 2020-07-10
Use fixed-point literal

turbolent:
Noticed this while running https://github.com/onflow/cadence/pull/226


joshuahannan:
I've fixed this in a big PR I'm working on so I'll close this. Thanks for pointing it out though!

turbolent:
OK, I guess I'll wait with the compatibility suite then – Enabling it right now would mean that it would fail because of this issue.

joshuahannan:
@turbolent The PR is here if you'd like to review

PR #30, opened 2020-07-10
Update template transactions and use them in templates package

joshuahannan:
* updated the transaction in the `transactions` directory so they were the accurate transactions for the fungible token
* Used those transaction in the `templates` package which is what the tests use. Now there is less copying and pasting.
* Changed the transactions and scripts to use arguments
* Changed the scripts to have return values which are checked in the tests 
* Updated dependencies


lib/go/test/token_test.go
```diff
@@ -105,21 +146,34 @@ func TestExternalTransfers(t *testing.T) {
...
 		tx := flow.NewTransaction().
 			SetScript(templates.GenerateTransferVaultScript(fungibleAddr, exampleTokenAddr, "ExampleToken")).
-			AddArgument(cadence.UFix64(30000_00000000)).
-			AddArgument(cadence.NewAddress(joshAddress)).
 			SetGasLimit(100).
 			SetProposalKey(b.ServiceKey().Address, b.ServiceKey().ID, b.ServiceKey().SequenceNumber).
 			SetPayer(b.ServiceKey().Address).
 			AddAuthorizer(exampleTokenAddr)
 
+		_ = tx.AddArgument(cadence.UFix64(30000_00000000))
```
turbolent, line 174: Consider using `NewUFix64("30000.0")` here and in other parts of the code. 
Using a literal like `30000_00000000` is brittle, because it's hard to see if it uses the correct amount of fractional digits, and there is no padding if it has not.

lib/go/test/token_test.go
```diff
@@ -90,13 +130,14 @@ func TestExternalTransfers(t *testing.T) {
...
 		tx := flow.NewTransaction().
 			SetScript(templates.GenerateTransferVaultScript(fungibleAddr, exampleTokenAddr, "ExampleToken")).
-			AddArgument(cadence.UFix64(0_00000000)).
-			AddArgument(cadence.NewAddress(joshAddress)).
 			SetGasLimit(100).
 			SetProposalKey(b.ServiceKey().Address, b.ServiceKey().ID, b.ServiceKey().SequenceNumber).
 			SetPayer(b.ServiceKey().Address).
 			AddAuthorizer(exampleTokenAddr)
 
+		_ = tx.AddArgument(cadence.UFix64(0_00000000))
```
turbolent, line 138: Maybe check the errors here and in other places of the codebase

transactions/burn_tokens.cdc
```diff
@@ -4,10 +4,10 @@
 //
 // The burning amount would be a parameter to the transaction
 
-import FungibleToken from 0xee82856bf20e2aa6
-import ExampleToken from 0x03
+import FungibleToken from 0xFUNGIBLETOKENADDRESS
+import ExampleToken from 0xTOKENADDRESS
 
-transaction {
+transaction(amount: UFix64) {
```
turbolent, line 10: 👍

PR #35, opened 2020-07-21
adding a cadence minifier

vishalchangrani:
For the Ledger integration, we need to provide a minified transaction which can be included in the ledger software that runs on the device. It was suggested that this could be used as an opportunity to also kick-off a little `candence` minifier 🙂 

If this is not the right home for such a code please lemme know and I will move it.


lib/go/minifier/minifier_test.go
```diff
@@ -0,0 +1,42 @@
...
+let recipient = getAccount(to)
+let receiverRef = recipient.getCapability(/public/exampleTokenReceiver)!.borrow<&{FungibleToken.Receiver}>()
+?? panic("Could not borrow receiver reference to the recipient's Vault")
+receiverRef.deposit(from: <-self.sentVault)
+}
+}
+`
+// Test to test the minifier function
+func TestMinify(t *testing.T) {
+	inputFile := "../../../transactions/transfer_tokens.cdc"
```
vishalchangrani, line 32: testing with the transaction that ledger will be using.

lib/go/minifier/minifier_test.go
```diff
@@ -0,0 +1,42 @@
...
+	outputFile, err := ioutil.TempFile("/tmp", "minified*.cdc")
```
turbolent, line 33: `/tmp` is not the temporary directory on all platforms. Consider using an empty string for the directory. From https://golang.org/pkg/io/ioutil/#TempDir:

> If dir is the empty string, TempDir uses the default directory for temporary files (see os.TempDir)

```suggestion
	outputFile, err := ioutil.TempFile("", "minified*.cdc")
```

lib/go/minifier/minifier_test.go
```diff
@@ -0,0 +1,42 @@
...
+	require.NoError(t, err)
+	fmt.Println(outputFile.Name())
```
turbolent, line 35:
```suggestion
```

lib/go/minifier/minifier_test.go
```diff
@@ -0,0 +1,42 @@
...
+	defer os.Remove(outputFile.Name())
```
turbolent, line 36: Consider introducing a variable for the repeated `outputFile.Name()`

lib/go/minifier/minifier.go
```diff
@@ -0,0 +1,79 @@
...
+				return nil
+			}
+			return err
+		}
+
+		line = strings.TrimSpace(line)
+		if line == "" || strings.HasPrefix(line, commentsPrefix){
+			continue
+		}
+		_, err = writer.WriteString(line + "\n")
```
turbolent, line 69:
```suggestion
		_, err = writer.WriteString(line)
		if err != nil {
			return err
		}

		_, err = writer.WriteRune('\n')
		if err != nil {
			return err
		}
```

lib/go/minifier/minifier.go
```diff
@@ -0,0 +1,79 @@
...
+	log.Println("output file:", *outputFile)
+
+	err := minify(*inputFile, *outputFile)
+	if err != nil {
+		log.Fatalf("failed to minify %s", *inputFile)
+	}
+
+	log.Println("done")
+}
+func minify(inputFile, outputFile string) error {
```
turbolent, line 40:
```suggestion
}

func minify(inputFile, outputFile string) error {
```

lib/go/minifier/minifier.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"bufio"
+	"flag"
+	"io"
+	"log"
+	"os"
+	"strings"
+)
+
+const commentsPrefix  = "//"
```
turbolent, line 12: Consider running gofmt on all files

```suggestion
const commentsPrefix = "//"
```

lib/go/minifier/minifier.go
```diff
@@ -0,0 +1,79 @@
...
+		line, err := reader.ReadString('\n')
+		if err != nil {
+			if err == io.EOF {
+				return nil
+			}
+			return err
+		}
+
+		line = strings.TrimSpace(line)
+		if line == "" || strings.HasPrefix(line, commentsPrefix){
```
turbolent, line 66:
```suggestion
		if line == "" || strings.HasPrefix(line, commentsPrefix) {
```

lib/go/minifier/go.mod
```diff
@@ -0,0 +1,7 @@
+module github.com/onflow/flow-ft/lib/go/minifier
+
+go 1.14
+
+require github.com/stretchr/testify v1.5.1
+
+replace github.com/onflow/flow-ft/lib/go/minifier => ../minifier
```
turbolent, line 7: Why is this needed? It seems to replace itself?

vishalchangrani: all `go.mod` have them e.g https://github.com/onflow/flow-ft/blob/d7d7a0d808b32481f96b34d4c892c184e071c9e1/lib/go/contracts/go.mod#L12. So I followed convention. I believe its for go.mod to use local copy if `minifier` is imported locally.

vishalchangrani:
> Nice!
> 
> If needed, we can work on something more sophisticated. This is highly related to pretty printing ([onflow/cadence#209](https://github.com/onflow/cadence/issues/209)), which we could use to generate minimal output. Furthermore, we could perform renaming of parameters and locals.
> 
> Not quite sure, but I feel like such a tool should probably just live in Cadence repo, given that it is not specific to the Fungible Token contract.

thanks for the review. I will move it over to the Cadence repo.

vishalchangrani:
Moved this change to the Cadence repo as per review comment (https://github.com/onflow/cadence/pull/252)

vishalchangrani:
closing this PR as this is now here https://github.com/onflow/cadence/pull/252

PR #39, opened 2020-09-08
Remove precondition in deposit function

turbolent:
@joshuahannan asked me to remove the precondition in the deposit function that the deposited amount must be larger than zero, as it is unnecessarily requires call-sites to handle the case specifically.
Removing the precondition is safe, as the type is unsigned, so cannot be negative.


joshuahannan:
Did you run the tests with the new contract? Need to run `make generate` in the `lib/go/contracts` directory to generate the contracts before testing.

turbolent:
No, I didn't realize that was necessary. I generated and pushed.

Tests are passing locally:

```
$ go test ./...
ok  	github.com/onflow/flow-ft/lib/go/contracts	0.184s
```

joshuahannan (approved):

PR #42, opened 2020-10-11
Update addresses

MaxStalker:
Update addresses for deployed FungibleToken definitions on testnet and mainnet


joshuahannan (approved):
Thank you so much!

PR #43, opened 2020-11-25
Use open-sourced emulator, enable tests on CI

turbolent:


lib/go/test/token_test.go
```diff
@@ -3,7 +3,8 @@ package test
 import (
 	"testing"
 
-	emulator "github.com/dapperlabs/flow-emulator"
+	emulator "github.com/onflow/flow-emulator"
+	templates2 "github.com/onflow/flow-go-sdk/templates"
```
joshuahannan, line 7:
```suggestion
	sdktemplates "github.com/onflow/flow-go-sdk/templates"
```

Maybe a more descriptive name here?

joshuahannan (approved):
So happy the emulator is open source now. Thanks for getting these updated! 😃

lib/go/test/token_test.go
turbolent: Good idea, improved it in 741ffbe6e5cffb266b01063f32fc23d423a928ed

PR #45, opened 2020-12-14
Fix asset path prefix stripping

turbolent:
go-bindata's `-prefix` flag takes a path prefix, but the argument for the flag ended with three dots, which is not a prefix for the paths.


joshuahannan (approved):
Thanks for cleaning this up!

PR #46, opened 2021-01-12
Update to Cadence v0.12.1, Go SDK v0.14.0, and Emulator v0.14.0-beta1

turbolent:
- Update dependencies: Cadence v0.12.1, Go SDK v0.14.0, and Emulator v0.14.0-beta1
- Update contracts, transactions, and scripts to use new Cadence features:
  - `getCapability` does not return an optional anymore
  - Use `Capability.check` instead of `Capability.borrow`
- Improve error pretty printing
- Improve the tests:
  - Fix the argument order of equality assertions
  - Refactor partial expression in long lines into separate variable declarations
  - Add line breaks to avoid long lines


joshuahannan (approved):
Thank you for adding this!

turbolent:
Thank you for reviewing @joshuahannan! I just updated the contracts, transactions, and scripts, could you please have another look at those? Thank you

joshuahannan (approved):
👍

PR #47, opened 2021-01-20
Adds private token receiver

joshuahannan:
* Adds a contract that manages a public token receiver that is only callable by an account with a special resource: `PrivateReceiverForwarder.cdc`

* Transaction to add that capability to an account: `create_private_forwarder.cdc`

* Transaction to setup an account with a vault, balance capability, and private receiver: `setup_and_create_forwarder.cdc`
* Transaction for sender to send tokens via the private receiver to many accounts: `transfer_private_many_accounts.cdc`
* Transaction for sender to send tokens normally to many accounts: `transfer_many_accounts.cdc`
* Transaction to create an account and use the private forwarder instead of normal receiver


transactions/transfer_many_accounts.cdc
```diff
@@ -0,0 +1,36 @@
+import FungibleToken from 0xFUNGIBLETOKENADDRESS
+import ExampleToken from 0xTOKENADDRESS
+
+transaction(amount: UFix64, to: [Address]) {
```
janezpodhostnik, line 4: In  [the minimum account "migration"](https://github.com/dapperlabs/flow-internal/issues/1354) we will be depositing a different amount to each account, so the input of the method could actually be: `addressAmountMap: {Address: UFix64}`. the same goes for the private transfer.

transactions/privateForwarder/create_private_forwarder.cdc
```diff
@@ -0,0 +1,19 @@
+import FungibleToken from 0xFUNGIBLETOKENADDRESS
+import ExampleToken from 0xTOKENADDRESS
+import PrivateReceiverForwarder from 0xPRIVATEFORWARDINGADDRESS
+
+transaction(receiver: Address) {
+
+    prepare(acct: AuthAccount) {
+        let recipient = getAccount(receiver)
+            .getCapability<&{FungibleToken.Receiver}>(/public/exampleTokenReceiver)
```
janezpodhostnik, line 9: I'm not sure I understand this part.

The idea is the recipient (of the FLOW) should only be able to receive FLOW from the private receiver. Which means it does not have any other `FungibleToken.Receiver`s. It only has a Vault in storage and a balance...

Maybe I'm reading something wrong.

transactions/privateForwarder/create_private_forwarder.cdc
```diff
@@ -0,0 +1,26 @@
+import FungibleToken from 0xFUNGIBLETOKENADDRESS
+import ExampleToken from 0xTOKENADDRESS
+import PrivateReceiverForwarder from 0xPRIVATEFORWARDINGADDRESS
+
+// This transaction creates a new private receiver in an account that
```
janezpodhostnik, line 5: I'm not 100% sure but I think dapper account creation also removes the balance capability. Should we account for that here?

transactions/privateForwarder/deploy_forwarder_contract.cdc
```diff
@@ -0,0 +1,9 @@
+transaction(contractName: String, code: [UInt8], senderStoragePath: StoragePath, storagePath: StoragePath, publicPath: PublicPath) {
+
+  prepare(signer: AuthAccount) {
+
+    signer.contracts.add(name: contractName, code: code, senderStoragePath, storagePath, publicPath)
```
janezpodhostnik, line 5: This looks like a standard contract deploy TX. How come it is in privateForwarder?

transactions/privateForwarder/create_private_forwarder.cdc
joshuahannan: Don't we need the balance capability to check storage capacity?

transactions/privateForwarder/deploy_forwarder_contract.cdc
joshuahannan: because we use it to deploy the forwarder contract and it uses arguments for the forwarder contract. Is there a more generic way to pass arguments to the generic create account functionality in the Go SDK?

transactions/privateForwarder/create_private_forwarder.cdc
janezpodhostnik: Yes, so it would need to be added back to dapper accounts if they are removing it.

joshuahannan: oh I see, I thought you were saying we should remove it. I'll add a line to add it back

joshuahannan: Actually, the transaction that we would be using is `setup_and_create_forwarder.cdc` so we don't need it in this one. This one is just to create one on a normal account

transactions/privateForwarder/deploy_forwarder_contract.cdc
janezpodhostnik: Ok, I see! Missed that part. Yeah I guess there is no template like that yet.

janezpodhostnik (approved):

PR #50, opened 2021-03-25
Setup and create private forwarder should be nilpotent

janezpodhostnik:
Setup and create private forwarder should be sate to run on the same account multiple times.


joshuahannan (approved):
Looks good!

PR #51, opened 2021-04-07
Update fungible token interface to check the type deposited in the deposit function

joshuahannan:


contracts/FungibleToken.cdc
```diff
@@ -182,6 +182,12 @@ pub contract interface FungibleToken {
         /// deposit takes a Vault and adds its balance to the balance of this Vault
         ///
         pub fun deposit(from: @Vault) {
+            // Assert that the concrete type of the deposited vault is the same
+            // as the vault that is accepting the deposit
+            pre {
+                from.getType().identifier.slice(from: 0, upTo: 18) == self.getType().identifier.slice(from: 0, upTo: 18): 
+                    "Cannot deposit an incompatible token type"
```
turbolent, line 189: This code does not match the comment, the code checks that the types are declared in the same account. Was that what you wanted to ensure?

If you really want to check that the type of the vault passes as an argument matches that of the vault that is accepting the deposit, you can just use `self.isInstance` and `from.getType()`:

```suggestion
                from.isInstance(self.getType()): 
                    "Cannot deposit an incompatible token type"
```

I've added a test case for this in onflow/cadence#779, see https://github.com/onflow/cadence/commit/4f93166492e3cdfb3c063c39a60d81d29570f9f9

It should also be possible to just compare the types: `from.getType() == self.getType()`

turbolent:
@joshuahannan Updated the comment above

contracts/FungibleToken.cdc
joshuahannan: oh yeah that makes a lot more sense! I'll change it to that. Thanks for the suggestion!

joshuahannan: Fixed in c84d060

joshuahannan:
@turbolent Does it look ok now?

turbolent (approved):
Looks good!

What is the plan for releasing this? This change might break existing contracts, though to be fair, it should be rare and likely already a bug (forgetting to cast, accepting something else)

joshuahannan:
I'd like to upgrade it soon. I don't think it'll break any existing contracts because they should have already been doing the force cast anyway. But we might need to have a conversation with some other stakeholders to make sure this is ok.

@robmyers @dete Are you ok with this change to the interface?

dete:
Yup! Seems like a correct addition to me.

bluesign:
does this not break forwarders?

joshuahannan:
@bluesign yes, I think you're correct, though I am surprised that the tests didn't pick this up. The pre-condition should be in the `Vault` object instead of the `Receiver` interface. I'll update that

bluesign:
@joshuahannan actually you have the correct implementation it seems. I had a total different imagination about forwarders, I was thinking they were implementing Vaults.

joshuahannan:
yeah you're right. I thought I had put it in the receiver interface, but I had actually put it in the Vault definition, so we're all good.

robmyers:
This is a nice use of modern Cadence rtti. 👍

PR #52, opened 2021-04-17
Update SECURITY.md

brianastrove:


brianastrove:
The update I made is to the reward part, min = 2500, up to 100k.

Also, question, is there a better email to send disclosures to than security@dapperlabs.com?

joshuahannan:
This looks good, but this security file is in almost every other onflow repo. Are you planning on updating those also?

joshuahannan (approved):

PR #53, opened 2021-06-03
Update to Cadence v0.16.1

turbolent:
Update to the latest version of Cadence. 

This uncovered an invalid test case: The private receiver forwarder contract used to set up a forwarder between different fungible token contracts, which is not supported: the test was forwarding from the newly deployed FT contract to the Flow token vault which is created by the Emulator and uses the bootstrapped FT contract. This used to work due to a bug


joshuahannan (approved):
Thanks! 👍

PR #68, opened 2022-05-24
post conditions and prepare-execute

joshuahannan:
Adds post-conditions to some of the transactions
Splits some of the transactions into prepare and execute blocks
adds path fields for the example token contract
adds some comments


transactions/mint_tokens.cdc
```diff
@@ -13,14 +13,19 @@ transaction(recipient: Address, amount: UFix64) {
     /// Reference to the Fungible Token Receiver of the recipient
     let tokenReceiver: &{FungibleToken.Receiver}
 
+    /// The total supply of tokens before the burn
+    let supplyBefore: UFix64
+
```
satyamakgec, line 18: We may add the `pre` condition as well to not allow the minting of `0` amount.

satyamakgec (approved):
LGTM

transactions/mint_tokens.cdc
joshuahannan: I don't like having that pre-condition any more, on contract methods and in transactions. it wouldn't make a difference here anyway

Issue #66, opened 2022-05-19
Build core FungibleTokenSwitchboard features


alilloig:
### Issue To Be Solved
Build the core features specified in the [epic](https://github.com/onflow/flow-ft/issues/62). Also build the tests for the features

### Suggest A Solution
* We will write a new contract “FungibleTokenSwitchboard” that will feature a resource “Switchboard” conforming to {FungibleToken.Receiver} but instead of storing a vault balance, it will store a record of these capabilities: Capability<&AnyResource{FungibleToken.Receiver}>.

* The deposit function enforced by the {FungibleToken.Receiver} will then check the type of the @FungibleToken.Vault received as an argument and route it to the proper Capability<&AnyResource{FungibleToken.Receiver}>. If there is no capability matching the deposited FT, the transaction will fail.

* The owner of the Switchboard resource will be able to call the function addVaultCapability(switch: Capability<&{FungibleToken.Receiver}> in order to store a new one of these capabilities.

* The Switchboard resource also should expose a getVaultCapabilities(): [Capability<&{FungibleToken.Receiver}>] in order to allow buyers to know if the token they are willing to use for the payment is supported by the seller / payment receiver.

Issue #65, opened 2022-05-09
Create Design Proposal for FT Switchboard


joshuahannan:
This issue is part of #62 

### Issue To Be Solved
For the fungible token switchboard contract, a design proposal document should be put together with the plan for the smart contract

### Suggest A Solution
Create a design document similar to a FLIP document with motivation, design summary, tradeoffs, docs plans, etc

[Draft proposal](https://www.notion.so/dapperlabs/Fungible-Token-Switchboard-Contract-Design-Document-e0183d982c5f4e52998437740fb189ab) (will be posted in #62 when complete)

PR #71, opened 2022-06-02
Fungible Token Switchboard contract and tests

alilloig:
Closes: #66 

## Description

Created FungibleTokenSwitchboard contract. Created transactions for setting up switchboard and manage them. Created test suite using js. Design proposal can be found [here](https://github.com/onflow/flow-ft/issues/65)

______


joshuahannan:
Can you put the `test` directory into a `lib/js/` directory so it isn't at the top level? Thank you!

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,151 @@
...
+///
+/// The contract that allows an account to receive payments in multiple fungible
+/// tokens using a single {FungibleToken.Receiver}
+///
+pub contract FungibleTokenSwitchboard {
+    
+    // Storage and Public Paths
+    pub let SwitchboardStoragePath: StoragePath
+    pub let SwitchboardPublicPath: PublicPath
+    pub let SwitchboardReceiverPublicPath: PublicPath
```
joshuahannan, line 13: I would maybe remove the `Switchboard` from these, so the full name doesn't get to be too much of a mouthful: `FungibleTokenSwitchboard.SwitchboardStoragePath` is a little bit redundant. :)

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,151 @@
...
+    
+    /// FungibleTokenSwitchboardInitialized
+    ///
+    /// The event emitted when the contract is created
+    ///
+    pub event FungibleTokenSwitchboardInitialized()
```
joshuahannan, line 19: probably don't need this

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,151 @@
...
+    /// The event that is emited when a new Switchboard resource is created
+    ///
+    pub event SwitchboardInitialized(switchboardResourceID: UInt64)
+
+    /// VaultCapacityAdded
+    ///
+    /// The event that is emitted when a new vault capacity is added to a
+    /// switchboard resource
+    ///
+    pub event VaultCapabilityAdded(type: Type)
```
joshuahannan, line 32: For all the events, I would think about what additional information might be useful to include. A good rule of thumb is to try to include any information that might be useful in events just in case people need them. For example, might be good to add an `address` field here and in the other ones. there may be more I'm not thinking of

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,151 @@
...
+    /// deposit method to deposit funds on it
+    ///
+    pub resource interface SwitchboardPublic {
+        pub fun getVaultCapabilities(): {Type: Capability<&{FungibleToken.Receiver}>}
+        pub fun deposit(from: @FungibleToken.Vault)
+    }
+    
+    /// Switchboard
+    /// The resource that stores the multiple fungible token receiver capabilities,
+    /// allowing the owner to add and remove then and anyone to deposit any
```
joshuahannan, line 54:
```suggestion
    /// allowing the owner to add and remove them and anyone to deposit any
```

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,151 @@
...
+    /// fungible token among the available capabilities
+    ///
+    pub resource Switchboard: FungibleToken.Receiver, SwitchboardPublic {
+        
+        /// fungibleTokenReceiverCapabilities
+        /// Dictoionary holding the fungible token receiver capabilities,
```
joshuahannan, line 60:
```suggestion
        /// Dictionary holding the fungible token receiver capabilities, 
```

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,151 @@
...
+        ///                    to the switchboard resource
+        ///            
+        /// 
+        /// Parameters: capability: The capability to expose a certain fungible
+        /// token vault deposit function through {FungibleToken.Receiver} that
+        /// will be added to the switchboard
+        ///
+        pub fun addVaultCapability(capability: Capability<&{FungibleToken.Receiver}>) {
+            // Borrow a reference to the vault pointed by the capability we want
+            // store inside the switchboard
```
joshuahannan, line 75:
```suggestion
            // to store inside the switchboard
```

transactions/switchboard/add_vault_capability.cdc
```diff
@@ -0,0 +1,27 @@
...
+      self.exampleTokenVaultCapabilty = signer.getCapability<&{FungibleToken.Receiver}>(ExampleToken.ReceiverPublicPath)
+      // Get a reference to the signers switchboard
+      self.switchboardRef = signer.borrow<&FungibleTokenSwitchboard.Switchboard>
+        (from: FungibleTokenSwitchboard.SwitchboardStoragePath) ?? panic("Could not borrow reference to switchboard")
+    }
+
+    execute {
+      // Add the capability to the switchboard using addVaultCapability method
+      self.switchboardRef.addVaultCapability(capability: self.exampleTokenVaultCapabilty)
+    }
```
satyamakgec, line 25: maybe good to add the post checks?

test/fungibletoken.test.js
```diff
@@ -0,0 +1,360 @@
...
+      sendTransaction({
+        name: "switchboard/setup_account",
+        args: [],
+        signers: [fungibleTokenSwitchboardUser]
+      })
+    );
+  });
+
+  // Second test checks if switchboard user is able to add ft token vault capabilities
+  test("should be able to create and remove vault capability", async () => {
```
satyamakgec, line 207: I think need to change the test statement as it is not removing the capability it is only adding.

test/fungibletoken.test.js
```diff
@@ -0,0 +1,360 @@
...
+    );
+    // Fourth step: verify that capabilities are returned
+    const [result, e] = await executeScript({
+      code: get_vault_capabilities,
+      args: [fungibleTokenSwitchboardUser]
+    });
+    expect(result).not.toBe(null);
+  });
+  
+});
```
satyamakgec, line 355: Maybe add a bad test as well to validate that contract would revert the transaction when non-added vault funds are transferred.

contracts/FungibleTokenSwitchboard.cdc
alilloig: Yeah totally when using those in another contract / tx the readability is awful.

contracts/FungibleTokenSwitchboard.cdc
alilloig: There is no real need of emitting an event every time a contract is deployed right?

contracts/FungibleTokenSwitchboard.cdc
alilloig: Great, I believe I have to learn about the real use case of consuming events, so far I have been coding them like mirroring for the examples but without having in mind how they will be used.

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,151 @@
...
+    pub fun createSwitchboard(): @Switchboard {
+        return <- create Switchboard()
+    }
+
+    init() {
+        self.SwitchboardStoragePath = StoragePath(identifier: "fungibleTokenSwitchboard")!
+        self.SwitchboardPublicPath = PublicPath(identifier: "fungibleTokenSwitchboardPublic")!
+        self.SwitchboardReceiverPublicPath = PublicPath(identifier: "GenericFTReceiver")!
+        // Emit the event that shows that the contract was initialized
+        emit FungibleTokenSwitchboardInitialized()
```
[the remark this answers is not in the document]
alilloig: just the event emit right?

contracts/FungibleTokenSwitchboard.cdc
satyamakgec: Yeah, there is no need for that. Sometimes it can be useful as well but generally, it depends on how systems are designed, Ex - when you have an off-chain service that is monitoring the actions of a particular address and once it sees that initialized event so it can start watching the activities of the newly deployed contracts as well. 

When seeing that event it can initialize the tables and whatnot within the off-chain database to curate the activity for the newly deployed contract.

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,151 @@
...
+        /// token vault deposit function through {FungibleToken.Receiver} that
+        /// will be added to the switchboard
+        ///
+        pub fun addVaultCapability(capability: Capability<&{FungibleToken.Receiver}>) {
+            // Borrow a reference to the vault pointed by the capability we want
+            // store inside the switchboard
+            let vaultRef = capability.borrow() 
+                ?? panic ("Cannot borrow reference to vault from capability")
+            // Use the vault reference type as key for storing the capability
+            self.fungibleTokenReceiverCapabilities[vaultRef.getType()] = capability
```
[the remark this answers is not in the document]
alilloig: to my knowledge they will have to store it in a different public path, which will be weird but still possible. I don't recall anything stoping you to have a thousand capabilities all of them pointing to the same resource as long as you store each of them on a different path, which shouldn't really happen as long as users utilize the usual tx templates

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,151 @@
...
+    /// allowing the owner to add and remove then and anyone to deposit any
+    /// fungible token among the available capabilities
+    ///
+    pub resource Switchboard: FungibleToken.Receiver, SwitchboardPublic {
+        
+        /// fungibleTokenReceiverCapabilities
+        /// Dictoionary holding the fungible token receiver capabilities, 
+        /// indexed by the fungible token vault type
+        ///
+        pub var fungibleTokenReceiverCapabilities: {Type: Capability<&{FungibleToken.Receiver}>}
```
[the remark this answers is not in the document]
alilloig: Yeah I did that on purpose I hated to see every variable pub and some access(contract) in dictionarys / arrays, but I agree, since we have a getter there's no real need of having it public

test/fungibletoken.test.js
```diff
@@ -0,0 +1,360 @@
...
+        signers: [serviceAccount]
+      })
+    );
+    // Step 4: Transfer 50 tokens from account A to account B
+    await shallPass(
+      sendTransaction({
+        name: "transfer_tokens",
+        args: [50, exampleTokenUserB],
+        signers: [exampleTokenUserA]
+      })
```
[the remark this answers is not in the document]
alilloig: @satyamakgec that will be using the jest expect function to check for the result of the tx right?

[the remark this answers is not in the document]
satyamakgec: Yes absolutely.

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,143 @@
...
+
+        /// getVaultCabilities function for get to know which tokens a certain
+        /// switchboard resource is prepared to receive
+        ///
+        /// Returns: The dictionary of stored {FungibleToken.Receiver} 
+        /// capabilities
+        ///
+        pub fun getVaultCapabilities(): {Type: Capability<&{FungibleToken.Receiver}>} {
+            return self.receiverCapabilities
+        }
```
bluesign, line 116: Do we need to give out Capabilities as a map ? I feel like just returning `keys` from here will be more useful. 

This would remove need for a public receiver link also.

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,143 @@
...
+    ///
+    /// Function that allows to create a new blank switchboard. A user must call
+    /// this function and store the returned resource in their storage
+    ///
+    pub fun createSwitchboard(): @Switchboard {
+        return <- create Switchboard()
+    }
+
+    init() {
+        self.StoragePath = StoragePath(identifier: "fungibleTokenSwitchboard")!
```
bluesign, line 137: nitpick little but  isn't it better to use non-dynamic paths in this context ?

`self.StoragePath = /storage/fungibleTokenSwitchboard` etc

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,143 @@
...
+        pub fun getVaultCapabilities(): {Type: Capability<&{FungibleToken.Receiver}>} {
+            return self.receiverCapabilities
+        }
+
+        init() {
+            // Initialize the capabilities dictionary
+            self.receiverCapabilities = {}
+            // Emit the event that indicates that a new switchboard resource
+            // has been created
+            emit SwitchboardInitialized(switchboardResourceID: self.uuid)
```
bluesign, line 123: is this event has any use? 

I think technically it is good practice to avoid such events if only information we can share is `uuid`. But I may be missing something.

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,143 @@
...
+            emit VaultCapabilityRemoved(type: vaultRef.getType())            
+        }
+        
+        /// deposit Takes a fungible token vault and routes it to the proper
+        ///         fungible token receiver capability for depositing it
+        /// 
+        /// Parameters: from: The deposited fungible token vault resource
+        ///        
+        pub fun deposit(from: @FungibleToken.Vault) {
+            let depositedVaultCapability = self.receiverCapabilities[from.getType()] ??
```
bluesign, line 101: oh I am surprised how this key is working: `from.getType()` 

@joshuahannan you know better but aren't they are different types? One is restricted and One is Vault ?

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,143 @@
...
+    
+    /// SwitchboardPublic
+    ///
+    /// The interface that enforces the method to allow anyone to check on the
+    /// available capabilities of a switchboard resource and also exposes the 
+    /// deposit method to deposit funds on it
+    ///
+    pub resource interface SwitchboardPublic {
+        pub fun getVaultCapabilities(): {Type: Capability<&{FungibleToken.Receiver}>}
+        pub fun deposit(from: @FungibleToken.Vault)
```
bluesign, line 43: I would either:

- `remove` deposit from this interface
- or make `SwitchboardPublic` to implement `FungibleToken.Receiver`

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,143 @@
...
+        /// 
+        /// Parameters: from: The deposited fungible token vault resource
+        ///        
+        pub fun deposit(from: @FungibleToken.Vault) {
+            let depositedVaultCapability = self.receiverCapabilities[from.getType()] ?? 
+                panic ("The deposited vault is not available on this switchboard")
+            let vaultRef = depositedVaultCapability.borrow() ?? 
+                panic ("Can not borrow a reference to the the vault")
+            vaultRef.deposit(from: <- from)
+        }
```
bluesign, line 106: hmm to discuss:

- should we `panic` in deposit ?

It feels so wrong for me to panic here, though we will end up with resource otherwise.

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,143 @@
...
+        /// 
+        /// Parameters: capability: The capability to a fungible token vault 
+        ///                         to be removed from the switchboard
+        ///
+        pub fun removeVault(capability: Capability<&{FungibleToken.Receiver}>) {
+            // Borrow a reference to the vault pointed by the capability we want
+            // store inside the switchboard            
+            let vaultRef = capability.borrow() 
+                ?? panic ("Cannot borrow reference to vault from capability")
+            // Use the vault reference to find the capability to remove
```
bluesign, line 89: Why not use here just `Type` ? do we really need `Capability` ?

bluesign:
This is great contract by functionality, but I think terrible UX. 

I suggest to improve:

- Let's get rid of capability part  ( I mean user facing part )  RemoveVault can take a `Type` which  is more natural, AddVault can actually take a `Path`. ( We can leave the `advanced usage` case as `AddVaultCapability`, so advanced users can add vaults in other accounts etc ) 

- Let's add `createSwitchBoardFromPaths( [PublicPath] ): @SwitchBoard` helper function. So I can just initialize this contract with a single line of code. 

- Later when storage enumeration lands, we can even auto define Public Receivers on init. ( whatever FungibleToken they are ) So if I have Flow, FUSD, USDC they will be all automatically defined. 


I also have very strong feelings on panic on deposit. But don't have a good suggestion yet.

contracts/FungibleTokenSwitchboard.cdc
joshuahannan: yeah, returning just the keys would be better

contracts/FungibleTokenSwitchboard.cdc
joshuahannan: yeah, probably can remove this

contracts/FungibleTokenSwitchboard.cdc
joshuahannan: The reference acts as a way to access the underlying type, so any built-in functions like `getType()` would return the same thing as the underlying type would. I agree it does seem a little weird, but that is the intended behavior

contracts/FungibleTokenSwitchboard.cdc
joshuahannan: yeah, I think that `SwitchboardPublic` should implement the receiver interface. Good catch!

contracts/FungibleTokenSwitchboard.cdc
joshuahannan: Any ideas of what we could do with the resource otherwise? We can't return it because that would change the deposit function signature

joshuahannan:
@bluesign I like your idea to use `Path` to add the vault and initialize the switchboard. I think it would be good to include that. `createSwitchBoardFromPaths( [PublicPath] ): @SwitchBoard` would also need an `Address` argument, right? so it knows which account to get it from?

I don't agree about using `Type` instead. That would require more lines of code in the transaction, whereas right now you can just get your receiver capability from your account and just pass that in directly instead of having to get the type also.

contracts/FungibleTokenSwitchboard.cdc
alilloig: great I'll change the function so it returns the keys, thanks @bluesign! Do you think getVaultTypes would be an ok name for the function? 

But I don't really understand why this change will remove the need for the public capability, could you help me understanding it?

pinkizpunk:
Fuck offf

Stacey

contracts/FungibleTokenSwitchboard.cdc
alilloig: Honestly I just knew about the dynamic function for creating paths while coding this and wanted to try it, do you know if there's any specific advantages / disadvantages  for each way?

contracts/FungibleTokenSwitchboard.cdc
alilloig: Yeah, that's the reason I asked this on stackoverflow https://stackoverflow.com/questions/72408565/reference-types-in-cadence, I believe it was you the first to answer me!!

contracts/FungibleTokenSwitchboard.cdc
alilloig: I'm afraid that's not possible [yet](https://stackoverflow.com/questions/72408565/reference-types-in-cadence)

contracts/FungibleTokenSwitchboard.cdc
alilloig: Yeah, I totally agree that's not really providing any value

test/fungibletoken.test.js
alilloig: that will requiere to add a new example token "2" or some other token to the repo, but I really like the idea of checking that. @joshuahannan any thoughts?

contracts/FungibleTokenSwitchboard.cdc
bluesign: People can add `private` capability, and then it can leak. But with this way, capability never leaves the resource.

contracts/FungibleTokenSwitchboard.cdc
bluesign: as I said there is no disadvantage, just syntax nitpick.

contracts/FungibleTokenSwitchboard.cdc
bluesign: hey @alilloig , I remember that question, I mean it was reference vs nonReference getType comparison right? 

Actually it makes sense, but this one has got me thinking long. I think this one is wrong, but I need to ask Bastian maybe.

contracts/FungibleTokenSwitchboard.cdc
bluesign: how about we add a second function to public interface, something like ‘safeDeposit’? So if who is sending knows that we are switchboard, they can use safeDeposit (which returns NFT vault) 

this way we provide extended functionality depending on the client while preserving standard.

bluesign:
@joshuahannan yeah you are right, `Address` is needed there.  

Or we can make it: `AddVaults( [PublicPath] )`, `AddVault( PublicPath )` ( which we can call after we save the switchboard to storage, so we can eliminate the address part )  we can panic with pre conditions with self.owner=nil 

For Type part: it is little tricky for me:

`pub fun removeVault(capability: Capability<&{FungibleToken.Receiver}>) `

here when I look at the signature, I see it will remove some capability, but actually it is removing some Type ( as key of the dictionary ) But I don't have strong feelings for this one. I am a bit torn too.

contracts/FungibleTokenSwitchboard.cdc
alilloig: In which way do you think it's wrong? @turbolent can you take a look at it?

contracts/FungibleTokenSwitchboard.cdc
joshuahannan: oh yeah. good point. I think it is fine to leave it the way it is

contracts/FungibleTokenSwitchboard.cdc
joshuahannan: That is an interesting idea. @alilloig Do you want to try adding that?

contracts/FungibleTokenSwitchboard.cdc
alilloig: I pushed the commit for this change, @satyamakgec could you take a look at the test for it? I'm checking for the result not to be null `expect(result).not.toBe(null);` but I don't like it, I'd love to have a test that checks for a non empty array but I didn't find a way to do that

contracts/FungibleTokenSwitchboard.cdc
bluesign: yeah I asked him already, all good

contracts/FungibleTokenSwitchboard.cdc
alilloig: I'm not sure if I'm understanding it properly, are we talking about something like this?
```
        /// safeDeposit Takes a fungible token vault and returns a reference to 
        ///             to the proper Vault from the switchboard, allowing the 
        ///             depositor to use the reference for depositing the funds
        ///             
        /// Parameters: vaultType: The type of the ft vault that wants to be deposited
        ///
        /// Returns: A reference to the required fungible token vault
        pub fun safeDeposit(vaultType: Type): &FungibleToken.Vault {

        }

Comment (bluesign) [reply contracts/FungibleTokenSwitchboard.cdc:99]: Something like this: 

```
pub fun safeDeposit(from: @FungibleToken.Vault): @FungibleToken.Vault {
    if let depositedVaultCapability = self.receiverCapabilities[from.getType()] {
        if let vaultRef = depositedVaultCapability.borrow() {
             vaultRef.deposit(from: <- from.withdraw(amount: from.balance) )
        }
    }
    return <- from 
}
```

so if any reason it cannot deposit, it will return the Vault back

Comment (alilloig) [reply contracts/FungibleTokenSwitchboard.cdc:99]: oh okey great!! I was not seeing the whole picture, I believe the point is to give payers a way to use the switchboard, probably focussed on the paying royalties use case, a way to have the transaction completed even if any of the receiver's vault is no available, neat! I actually love it thank you @bluesign !!!

Comment (alilloig) [reply contracts/FungibleTokenSwitchboard.cdc:100]: check the commit and let me know what you think guys. I coded it with code readability in mind, let me know if you find it "over commented" but I wanted to emphasize and make cristal clear the difference between that and the regular deposit function

**Comment by alilloig** [comment]
For the type discussion I'm still aligned with the original implementation, I did it with the tx in mind, I believe it's the simpler way of storing it into the switchboard as josh said. Keep in mind @bluesign that we are actually deleting the dictionary entry that is storing the capability that you are passing as an argument, we are just using the Type of the reference that can be borrowed from that capability as the key for indexing that.

And relating to add new capabilities using paths, I also like the idea, but I believe we should pass an address also as an argument. The other option will be to check sor resource.owner, and even if the Switchboard resource probably is extremely unlikely to abandon their owner's account, it still feels a little like checking msg.sender, and that could be just avoided passing the Address parameter. Also we will be adding the feature of adding several vaults with just one call to the resource which I think is quite nice. What do you think of something like this?

```   /// addNewVaultsByPath adds a number of new fungible token receiver 
        ///                    capabilities by using the paths where they are
        ///                    stored
        ///
        /// Parameters: paths: The paths where the public capabilities are stored
        ///             address: The address of the owner of the capabilities
        ///
        pub fun addNewVaultsByPath(paths: [PublicPath], address: Address) {

File: contracts/FungibleTokenSwitchboard.cdc [hunk 21]
```diff
@@ -69,6 +69,31 @@ pub contract FungibleTokenSwitchboard {
...
+            // Get the account where the public capabilities are stored
+            let owner = getAccount(address)
+            // For each path, get the saved capability and store it 
+            // into the switchboard's receiver capabilities dictionary 
+            for path in paths {
+                let capability = owner.getCapability<&{FungibleToken.Receiver}>(path)
+                // Borrow a reference to the vault pointed by the capability we want
+                // to store inside the switchboard
+                let vaultRef = capability.borrow() 
+                    ?? panic ("Cannot borrow reference to vault from capability")
```
bluesign, line 89: Can we maybe make this panic free ? I had in my mind to keep somewhere list of common paths, and initialize with that.

bluesign:
> pub fun addNewVaultsByPath(paths: [PublicPath], address: Address) 

this looks ok to me, just commenting something small on implementation.

contracts/FungibleTokenSwitchboard.cdc
alilloig: I'm not 100% sure on what will be preferable for users, since if we make it panic free, users will need to check after which vaults where actually added. @joshuahannan what do you think about it?

bluesign: I mean with event I guess it is more manageable, otherwise event is useless as is, we need to remove that. But it would be nice to have a whitelist of paths ( for example at marketplace ) 

also I strongly favor not overwriting the capability. I mean if I already forwarding flow, I should remove to redirect it to another capability.

alilloig: You are absolutely right, using the event consumers could know which of their vaults were added to the switchboard. 

About overwriting that's something worth a little discussion, but probably not being able to overwrite it by mistake makes a lot of sense.

joshuahannan:
I think requiring them to delete it first before overwriting is a good idea

alilloig:
> I think requiring them to delete it first before overwriting is a good idea

Great, if it makes sense I'll make the `addNewVault()` panicking if there's already a vault for that type, while keeping `addNewVaultsByPath()` panic free

transactions/switchboard/transfer_tokens.cdc
```diff
@@ -0,0 +1,41 @@
...
+// switchboard and have add the proper capability to it
+//
+// The withdraw amount and the account from getAccount
+// would be the parameters to the transaction
+
+import FungibleToken from "./../../contracts/FungibleToken.cdc"
+import FungibleTokenSwitchboard from "./../../contracts/FungibleTokenSwitchboard.cdc"
+import ExampleToken from "./../../contracts/ExampleToken.cdc"
+
+transaction(to: Address, amount: UFix64) {
```
joshuahannan, line 13:
```suggestion
import ExampleToken from "./../../contracts/ExampleToken.cdc"

transaction(to: Address, amount: UFix64, receiverPath: PublicPath) {
```

instead of explicitly using the switchboard contract here for this normal transfer, can you just have the user pass in an argument that is the switchboard public path and just borrow the receiver from there?

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,208 @@
...
+        ///             the funds, avoiding panicking if the vault is not available
+        ///             
+        /// Parameters: vaultType: The type of the ft vault that wants to be 
+        ///                        deposited
+        ///
+        /// Returns: The deposited fungible token vault resource, without the
+        ///          funds if the deposit was succesful, or still containing the
+        ///          funds if the reference to the needed vault was not found
+        ///
+        pub fun safeDeposit(from: @FungibleToken.Vault): @FungibleToken.Vault {
```
joshuahannan, line 157: I think this should return an optional. Do you think that makes sense?

alilloig:
What do you think of an event that would be emitted if a deposit fails?

```cadence
    /// NotCompletedDeposit
    /// 
    /// The event that is emitted when a deposit can not be completed
    ///
    pub event NotCompletedDeposit(type: Type, amount: UFix64, 
                                    switchboardOwner: Address)

        /// deposit Takes a fungible token vault and routes it to the proper
        ///         fungible token receiver capability for depositing it
        /// 
        /// Parameters: from: The deposited fungible token vault resource
        ///        
        pub fun deposit(from: @FungibleToken.Vault) {
            if let depositedVaultCapability = self.receiverCapabilities[from.getType()]{
                if let vaultRef = depositedVaultCapability.borrow() {
                    vaultRef.deposit(from: <- from)
                } else {
                    emit NotCompletedDeposit(type: from.getType(), 
                                        amount: from.balance, 
                                        switchboardOwner: self.owner!.address)
                    panic ("Can not borrow a reference to the the vault")
                }
            } else {
                emit NotCompletedDeposit(type: from.getType(), 
                                        amount: from.balance, 
                                        switchboardOwner: self.owner!.address)
                panic ("The deposited vault is not available on this switchboard")
            }
        }


        /// safeDeposit Takes a fungible token vault and tries to route it to 
        ///             the proper fungible token receiver capability for 
        ///             depositing the funds, avoiding panicking if the vault is 
        ///             not available
        ///             
        /// Parameters: vaultType: The type of the ft vault that wants to be 
        ///                        deposited
        ///
        /// Returns: The deposited fungible token vault resource, without the
        ///          funds if the deposit was succesful, or still containing the
        ///          funds if the reference to the needed vault was not found
        ///
        pub fun safeDeposit(from: @FungibleToken.Vault): @FungibleToken.Vault? {
            // Try to get the proper vault capability from the switchboard
            // If the desired vault is present on the switchboard...
            if let depositedVaultCapability = self.receiverCapabilities[from.getType()] {
                // We try to borrow a reference to the vault from the capability
                // If we can borrow a reference to the vault...
                if let vaultRef =  depositedVaultCapability.borrow(){
                    // We deposit the funds on said vault
                    vaultRef.deposit(from: <- from)
                    return nil
                } else {                    
                    // If we can't borrow a reference to the vault we emit the
                    // proper event and return the passed vault so the depositer 
                    // can recover their funds
                    emit NotCompletedDeposit(type: from.getType(), 
                                        amount: from.balance, 
                                        switchboardOwner: self.owner!.address)

                    return <- from
                }
            } else {
                // If the vault was not found on the switchboard and the deposit
                // could not be completed we emit the proper event and return
                // the passed vault so the depositer can recover their funds
                emit NotCompletedDeposit(type: from.getType(), 
                                        amount: from.balance, 
                                        switchboardOwner: self.owner!.address)              
                return <- from
            }
        }
```

joshuahannan:
If the execution reverts, then the events aren't emitted I believe so it wouldn't make sense for the regular deposit. It would probably be good for the safe deposit though

bluesign:
great idea. Just code is little too complex now to grasp at first sight.  maybe something like this?

```cadence
 pub fun safeDeposit(from: @FungibleToken.Vault): @FungibleToken.Vault? {
            // Try to get the proper vault capability from the switchboard
            // If the desired vault is present on the switchboard...
            if let depositedVaultCapability = self.receiverCapabilities[from.getType()] {
                // We try to borrow a reference to the vault from the capability
                // If we can borrow a reference to the vault...
                if let vaultRef =  depositedVaultCapability.borrow(){
                    // We deposit the funds on said vault
                    vaultRef.deposit(from: <- from.withdraw(amount: from.balance))
                }
            }
            
            // if deposit failed for some reason 
            if from.balance>0{
                emit NotCompletedDeposit(type: from.getType(), 
                                        amount: from.balance, 
                                        switchboardOwner: self.owner!.address)              
                return <- from
            }
            
            destroy from 
            return nil
}
```

alilloig:
> If the execution reverts, then the events aren't emitted I believe so it wouldn't make sense for the regular deposit. It would probably be good for the safe deposit though

thank you, I wasn't 100% sure if the events emitted before the execution was aborted were emitted anyways

alilloig:
@bluesign I liked your way, event added!

alilloig:
And I came up with a thought, should not the `getVaultTypes()` method check if all the stored capabilities can be borrowed, and if any of them cannot be borrowed, don't return that Type? So the caller can be truly aware of which FT can actually deposit into the switchboard

alilloig:
as soon as the new version of the flow-js-testing is released on npm (it will add support for `Path`s) I'll update it and request a re-review from you guys

transactions/switchboard/add_vault_capability.cdc
```diff
@@ -0,0 +1,30 @@
...
+transaction {
+
+    let exampleTokenVaultCapabilty: Capability<&{FungibleToken.Receiver}>
+    let switchboardRef:  &FungibleTokenSwitchboard.Switchboard
+
+    prepare(signer: AuthAccount) {
+      // Get the example token vault capability from the signer's account
+      self.exampleTokenVaultCapabilty = 
+        signer.getCapability<&{FungibleToken.Receiver}>
+                                (ExampleToken.ReceiverPublicPath)
```
satyamakgec, line 18: I would recommend adding the check whether the capability exists because the failed transaction pollutes the blockchain and charges transaction costs. We can also save before submitting the transaction if we have a check in the transaction.  (If I am not wrong)

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -0,0 +1,237 @@
...
+            // For each path, get the saved capability and store it 
+            // into the switchboard's receiver capabilities dictionary 
+            for path in paths {
+                let capability = owner.getCapability<&{FungibleToken.Receiver}>(path)
+                // Borrow a reference to the vault pointed to by the capability we 
+                // want to store inside the switchboard
+                // If the vault was borrowed successfully...
+                if let vaultRef = capability.borrow() {
+                    // ...and there is no previous capability added for that token
+                    if (self.receiverCapabilities[vaultRef!.getType()] == nil){
```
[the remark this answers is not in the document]
satyamakgec: It is not a bad thing to have as it minimises the chances of error but I think for this condition, it is fine to have as it is.

joshuahannan:
@alilloig Can you pull the latest from master and fix the merge conflicts that are shown here? Thanks!

alilloig:
@joshuahannan I believe is ready to merge!

joshuahannan (approved):

pinkizpunk:
Fuck off

PR #79, opened 2022-07-21
add destroy conditional

joshuahannan:
Adds a conditional in the deposit block of `ExampleToken` so that the assignment only happens when it needs to


alilloig (approved):
I saw this on the new standards, does the calculation matter in terms of gas use?

joshuahannan:
It will save some gas, but the savings for an individual transaction are negligible, the improvement is more of an aggregate improvement over all transactions.

PR #80, opened 2022-07-22
Adds a generic transfer transaction

joshuahannan:
Adds a generic transfer transaction that allows the submitter to provide storage and public paths for any contract without having to import those fungible token contracts.


transactions/generic_transfer.cdc
```diff
@@ -0,0 +1,29 @@
+import FungibleToken from "./../contracts/FungibleToken.cdc"
+
+// Can pass in any storage path and receiver path instead of just the default.
+// This lets you choose the token you want to send as well the capability you want to send it to.
+transaction(amount: UFix64, to: Address, senderPath: StoragePath, receiverPath: PublicPath) {
```
satyamakgec, line 5: This is a risky transaction as it would allow fetching the funds from any FT. I would recommend adding a caution message at the top as it would allow the attacker to use an unwanted path for transfer that would lead to the transfer of unwanted tokens.

joshuahannan: I'm not very worried about that. There is always a risk of an app providing a malicious transaction, so it is the responsibility of the wallet to make sure they are checking the transaction for correctness and safety. I'll still put the comment though

alilloig: Also any decent ft implementation will check for that on the deposit function right? Actually I just re-visited the contract interface and we have a precon in the deposit function that won't allow any unwanted tokens if I'm not mistaken
```cadence
            pre {
            }
```

alilloig (approved):

transactions/generic_transfer.cdc
joshuahannan: He is referring to transferring any fungible token to the correct receiver. So a dapp provides me with this transaction saying that they are transferring FLOW, but they transfer USDC instead.

alilloig: oh I understand thx you!!

Issue #83, opened 2022-07-29
Add FT Metadata Views to Example Token contract


alilloig:
### Issue To Be Solved
We need to implement the FT Metadata Views on the `ExampleToken.cdc` contract so it can be used as an example on how to use them.

PR #89, opened 2022-09-01
Create FungibleTokenMetadataViews contract

alilloig:
Closes: #82 
## Description

Creates the contract for the Fungible Token's Metadata Views.

______

For contributor use:

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [x] Code follows the [standards mentioned here](https://github.com/onflow/flow-ft/blob/master/CONTRIBUTING.md#styleguides).
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels
- [x] Update the version in package.json when there is a change in the smart contracts

PR #91, opened 2022-09-15
Implement MetadataViews on ExampleToken

alilloig:
Closes: #83 

## Description

Implements `MetadataViews.getViews()` and `MetadataViews.resolveView()` in the `ExampleToken` contract featuring the new views defined in `FungibleTokenMetadataViews`

______

For contributor use:

- [x] Targeted PR against `alilloig/ft-metadata-views` branch


alilloig:
PR template is showing automatically when doing the PR from the gh web interface!! 🎉

alilloig:
I'll love to get some ideas about which test cases I can do here. Right now there's only one, setting up a vault using just the path to an existing vault in some other account using [this transaction](https://github.com/onflow/flow-ft/blob/alilloig/example-token-metadata-views/transactions/setup_account_from_vault_reference.cdc). Pretty cool to setup a vault without any reference to the token contract btw!!

contracts/ExampleToken.cdc
```diff
@@ -106,10 +89,69 @@ pub contract ExampleToken: FungibleToken {
...
+            publicLinkedType: Type<&{FungibleToken.Receiver, FungibleToken.Balance, MetadataViews.Resolver}>(),
+            providerLinkedType: Type<&ExampleToken.Vault{FungibleToken.Provider, MetadataViews.Resolver}>(),
+            createEmptyVaultFunction: (fun (): @ExampleToken.Vault {
+                return <-ExampleToken.createEmptyVault()
+            })
+        )
+    }
+
+    // Auxiliary function to return the FTDisplay view
+    access(contract) fun returnDisplayView(): FungibleTokenMetadataViews.FTDisplay {
```
alilloig, line 136: Would love to read some thoughts about this functions since this should be focused on "being documentation" and I'm not sure if it brings clarity on how to resolve views. Working on that I just noticed that we do not have an example on how to solve the `NFTView` in the `ExampleNFT` contract

contracts/ExampleToken.cdc
```diff
@@ -58,21 +45,20 @@ pub contract ExampleToken: FungibleToken {
...
     /// new tokens.
     ///
-    pub resource Vault: FungibleToken.Provider, FungibleToken.Receiver, FungibleToken.Balance {
+    pub resource Vault: FungibleToken.Provider, FungibleToken.Receiver, FungibleToken.Balance, MetadataViews.Resolver {
 
         /// The total balance of this vault
         pub var balance: UFix64
 
-        // initialize the balance at resource creation time
+        ///Do we need extra fields for metadata?
```
joshuahannan, line 53: Yes, we should add extra fields for the display view

contracts/ExampleToken.cdc
```diff
@@ -1,53 +1,40 @@
...
 
     /// Total supply of ExampleTokens in existence
     pub var totalSupply: UFix64
-
+    
     /// Storage and Public Paths
     pub let VaultStoragePath: StoragePath
     pub let ReceiverPublicPath: PublicPath
     pub let BalancePublicPath: PublicPath
+    pub let ResolverPublicPath: PublicPath
```
joshuahannan, line 14: Why define a new path for the resolver? You can just put the resolve view functions in the balance interface and add default implementations now

contracts/ExampleToken.cdc
```diff
@@ -106,10 +89,74 @@ pub contract ExampleToken: FungibleToken {
...
+                    return ExampleToken.returnDisplayView()
+                case Type<FungibleTokenMetadataViews.FTVaultData>():
+                    return ExampleToken.returnVaultDataView()
+            }
+            return nil
+        }
+    }
+
+    // Auxiliary function to return the FTVaultData view
+    access(contract) fun returnVaultDataView(): FungibleTokenMetadataViews.FTVaultData {
```
joshuahannan, line 126: Why do you want the view resolvers to be outside the Vault type? I don't necessarily think it is bad, I'm just curious to hear your reasoning. I would probably prefer to have them defined in the `Vault` type

contracts/FungibleTokenMetadataViews.cdc
```diff
@@ -0,0 +1,168 @@
...
+        init(
+            storagePath: StoragePath,
+            publicPath: PublicPath,
+            providerPath: PrivatePath,
+            publicLinkedType: Type,
+            providerLinkedType: Type,
+            createEmptyVaultFunction: ((): @FungibleToken.Vault)
+        ) {
+            pre {
+                publicLinkedType.isSubtype(of: Type<&{FungibleToken.Receiver, FungibleToken.Balance, MetadataViews.Resolver}>()): "Public type must include FungibleToken.Receiver, FungibleToken.Balance and MetadataViews.Resolver interfaces."
```
joshuahannan, line 142: I don't think this is going to work with the way we've designed the paths above. The balance and receiver interfaces are linked at different public paths, and currently, the resolver is also at a different path. I think that the resolver should be linked at the same path as balance and then we'll probably need to update this view to have two public paths (the metadata path and the receiver path). Make sure there is a test case for this too

joshuahannan (commented):
Like I said in one of my comments, can you also add a `MetadataViews.Resolver` conformance to the vault type in `FungibleToken` and add default implementations for getViews and resolveView to it? They'll just return and empty array and `nil`. Default implementations is going to be in the next Flow spork, so it would be good to have this ready to deploy with those.

contracts/ExampleToken.cdc
alilloig: That's how I was doing it at the beginning (actually I was using the receiver path) but just when writing the transactions it seemed to me that I'd be easy to understand how it worked with an extra, a little unnecessary,  path. 
Actually I just realised that I forgot to use it in the `init()` function of the contract
```cadence
        // Create a public capability to the stored Vault that exposes
        // the `deposit` method through the `Receiver` interface and also
        // the `resolveView`method through the MetadataViews `Resolver` interface
        self.account.link<&{FungibleToken.Receiver, MetadataViews.Resolver}>(
            self.ReceiverPublicPath,
            target: self.VaultStoragePath
        )
```
Any particular reason it would be better to use the balance path rather than the receiver path?

contracts/ExampleToken.cdc
alilloig: Just because when you return the `FTView` you also need to create a new `FTVaultData` and a new `FTDisplay` I kind of hate to see duplicate code

contracts/FungibleTokenMetadataViews.cdc
alilloig: Actually now there's only one test case using this transaction https://github.com/onflow/flow-ft/blob/alilloig/example-token-metadata-views/transactions/setup_account_from_vault_reference.cdc

I just realise I should extend the case a little bit more and try to actually use the freshly new created `Vault`, but basically `<&{FungibleToken.Receiver, FungibleToken.Balance, MetadataViews.Resolver}>` are exposed at `/public/exampleTokenReceiver`.

I should improve consistency around how the paths are used probably.

contracts/ExampleToken.cdc
joshuahannan: Balance is a piece of metadata, so it makes sense that it would be the one that has the rest of the metadata.

contracts/FungibleTokenMetadataViews.cdc
joshuahannan: yeah, I would maybe expose all the interfaces at the receiver path and the balance path just to be thorough

contracts/ExampleToken.cdc
alilloig: what would you think about renaming the `ExampleToken.BalancePublicPath` as `ExampleToken.MetadataPublicPath` and expose both Balance and Resolver there?

contracts/ExampleToken.cdc
alilloig: Yep fixed, I'm kind of embarrassed I didn't think about that 🤦

contracts/FungibleTokenMetadataViews.cdc
alilloig: I refurbished everything about the public paths on the `FTVaultData` view, you can see it here https://github.com/onflow/flow-ft/pull/89/commits/089b9fbd31d63bd09ed7d10cfd5af5dc9f0e5364 and it has already been updated for using it that way on the ExampleToken

alilloig: So I didn't expose all interfaces on both paths (that could be done easily) but create two separate fields on the `FTVaultData` one for the receiver path and other for the metadata path, with its correspondent linked types

contracts/ExampleToken.cdc
alilloig: Do you mean including in the example token contract fields for its symbol or description?

Or do you mean to include the `MetadataViews.Display` view as one of the views that the vault resolves?

contracts/ExampleToken.cdc
```diff
@@ -106,10 +88,66 @@ pub contract ExampleToken: FungibleToken {
...
+        /// @param view: The Type of the desired view.
+        /// @return A structure representing the requested view.
+        ///
+        pub fun resolveView(_ view: Type): AnyStruct? {
+            switch view {
+                case Type<FungibleTokenMetadataViews.FTView>():
+                    return FungibleTokenMetadataViews.FTView(
+                        ftDisplay: self.resolveView(Type<FungibleTokenMetadataViews.FTDisplay>()) as! FungibleTokenMetadataViews.FTDisplay?,
+                        ftVaultData: self.resolveView(Type<FungibleTokenMetadataViews.FTVaultData>()) as! FungibleTokenMetadataViews.FTVaultData?
+                    )
```
psiemens, line 114: I find it easier to read the `resolveView` function when it uses helpers for each view type.

```cdc
pub fun resolveView(_ view: Type): AnyStruct? {
  switch view {
    case Type<FungibleTokenMetadataViews.FTView>():
      return self.resolveFTView()
    case Type<FungibleTokenMetadataViews.FTDisplay>():
      return self.resolveFTDisplayView()
      
    // ...
  }
}

// You can then avoid those clunky cast operations when resolving sub-views
pub fun resolveFTView(): FungibleTokenMetadataViews.FTView {
  return FungibleTokenMetadataViews.FTView(
    // Note: Cadence will automatically coerce FungibleTokenMetadataViews.FTView
    // into the optional FungibleTokenMetadataViews.FTView?
    ftDisplay: self.resolveFTDisplayView(),
    ftVaultData: self.resolveFTVaultDataView()
  )
}

// ...
```

contracts/ExampleToken.cdc
```diff
@@ -193,38 +219,35 @@ pub contract ExampleToken: FungibleToken {
...
 
         let admin <- create Administrator()
         self.account.save(<-admin, to: self.AdminStoragePath)
 
         // Emit an event that shows that the contract was initialized
-        //
         emit TokensInitialized(initialSupply: self.totalSupply)
     }
 }
+
```
psiemens, line 253: Extra line break here

contracts/FungibleTokenMetadataViews.cdc
```diff
@@ -17,13 +17,13 @@ pub contract FungibleTokenMetadataViews {
     ///
     pub struct FTView {
         pub let ftDisplay: FTDisplay?     
-        pub let vaultData: FTVaultData?
+        pub let ftVaultData: FTVaultData?
```
psiemens, line 20: 👍

contracts/ExampleToken.cdc
psiemens: I'm not sure how I feel about renaming the `BalancePublicPath` to `MetadataPublicPath`. Given that neither makes perfect sense, I'd be inclined to stick with the one that's already in use.

psiemens: If we are going to change the suggested name, something like `VaultPublicPath` makes more sense to me, since that can be used to describe the "publicly available information for this vault".

alilloig: I'd prefer `VaultPublicPath` over Balance if we are exposing also the Metadata there, so I'll go with that.

contracts/ExampleToken.cdc
alilloig: I learnt to code with C, I like my newlines at EOF jajaja. Deleted!

alilloig:
Thanks a lot for your comments @psiemens!! I've finally changed the vault's public path name to `VaultPublicPath`, I guess we all agree on that the "lesser bad" solution.

@satyamakgec @sisyphusSmiling I'd like to talk with you briefly about the tests to assure I'm not missing anything before merging this, I'll put some time on your calendars for Monday.

PR #94, opened 2022-09-27
Including the resource interfaces inside the `FungibleToken` contract interface.

alilloig:
## Description

This is more a code review/suggestion than a PR, but I thought it will be easier to read it this way.

This basically puts all the resource interfaces inside the FungibleToken contract interface so developers do not need to import two different contracts when creating FT. Only thing we are giving up is the implementation of the init function in the `VaultInfo` struct.


joshuahannan:
Thanks for the suggestion! We'll see what other people think before we include any of this

alilloig:
I'll flag this in the actual PR conversation if that makes sense!

PR #112, opened 2023-01-13
Make TokenForwarder and Switchboard  target capabilities retrievable

austinkline:
Closes: #???

## Description

Add a few helper methods that will make it easier to validate underlying receiver capabilities in the `TokenForwarder` and `FungibleTokenSwitchboard` contracts. Without this, it is much harder to know what we are looking at under the hood, which means that malicious users could setup any number of roundabout ways to make receivers actually invalid (like an infinite loop of receivers that will drain all gas in a transaction)

Made a brief diagram to showcase this problem, it applies to both TokenForwarding and the Switchboard:
flow-ft

In any case, without some mechanism to investigate what is being referenced in the forwarder or switchboard resources, there isn't a way on-chain to ensure that what we are sending through can actually reach its destination. Because of that, I can "poison" the chain of resources anywhere down the line to prevent a token vault from being able to be sent. 

With these editions, we can make some form of validator that only permits a maximum depth, and rejects the deposit if it goes beyond that. This will allow marketplaces to be defensive and reject vault distribution if it goes too far, or if it has the wrong type along its chain (a non forwarder or switchboard)
______

For contributor use:

### DIFF
--- a/contracts/utility/TokenForwarding.cdc
+++ b/contracts/utility/TokenForwarding.cdc
@@ -43,6 +43,14 @@ pub contract TokenForwarding {
             emit ForwardedDeposit(amount: balance, from: self.owner?.address)
         }
 
+        pub fun check(): Bool {
+            return self.recipient.check<&{FungibleToken.Receiver}>()
+        }
+
+        pub fun safeBorrow(): &{FungibleToken.Receiver}? {
+            return self.recipient.borrow<&{FungibleToken.Receiver}>()
+        }
+
         // changeRecipient changes the recipient of the forwarder to the provided recipient
         //
         pub fun changeRecipient(_ newRecipient: Capability) {
@@ -66,3 +74,4 @@ pub contract TokenForwarding {
         return <-create Forwarder(recipient: recipient)
     }
 }
+ 
--- a/contracts/utility/TokenForwarding.cdc
+++ b/contracts/utility/TokenForwarding.cdc
@@ -21,7 +21,12 @@ pub contract TokenForwarding {
     // Event that is emitted when tokens are deposited to the target receiver
     pub event ForwardedDeposit(amount: UFix64, from: Address?)
 
-    pub resource Forwarder: FungibleToken.Receiver {
+    pub resource interface ForwarderPublic {
+        pub fun check(): Bool
+        pub fun safeBorrow(): &{FungibleToken.Receiver}?
+    }
+
+    pub resource Forwarder: FungibleToken.Receiver, ForwarderPublic {
 
         // This is where the deposited tokens will be sent.
         // The type indicates that it is a reference to a receiver
@@ -74,4 +79,3 @@ pub contract TokenForwarding {
         return <-create Forwarder(recipient: recipient)
     }
 }
- 
--- a/contracts/FungibleTokenSwitchboard.cdc
+++ b/contracts/FungibleTokenSwitchboard.cdc
@@ -210,6 +210,35 @@ pub contract FungibleTokenSwitchboard {
             return nil
         }
 
+        /// Checks that the capability tied to a type is valid
+        ///
+        /// @param vaultType: The type of the ft vault whose capability needs to be checked
+        ///
+        /// @return a boolean marking the capability for a type as valid or not
+        pub fun checkReceiverByType(type: Type): Bool {
+            if self.receiverCapabilities[type] == nil {
+                return false
+            }
+
+            return self.receiverCapabilities[type]!.check()
+        }
+
+        /// Gets the receiver assigned to a provided vault type.
+        /// This is necessary because without it, it is not possible to look under the hood and see if a capability
+        /// is of an expected type or not. This helps guard against infinitely chained TokenForwarding or other invalid 
+        /// malicious kinds of updates that could prevent listings from being made that are valid on storefronts.
+        ///
+        /// @param vaultType: The type of the ft vault whose capability needs to be checked
+        ///
+        /// @return an optional receiver capability for consumers of the switchboard to check/validate on their own
+        pub fun safeBorrowByType(type: Type): &{FungibleToken.Receiver}? {
+            if !self.checkReceiverByType(type: type) {
+                return nil
+            }
+
+            return self.receiverCapabilities[type]!.borrow()
+        }
+
         /// A getter function to know which tokens a certain switchboard 
         /// resource is prepared to receive.
         ///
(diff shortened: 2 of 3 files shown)


austinkline:
Here is an example contract that could use these changes:
https://gist.github.com/austinkline/f7f80ec5ea04ef168de59a0778ddc438

**NOTE: This is not tested, it is meant to demonstrate the problem and why a solution like this is necessary**

joshuahannan (approved):
Really great suggestion, and well implemented. Are you able to add a test for these? I can try to get others to look at it

austinkline:
> Really great suggestion, and well implemented. Are you able to add a test for these? I can try to get others to look at it

Should have some time next week to add them 🙏

contracts/utility/TokenForwarding.cdc
```diff
@@ -43,6 +48,14 @@ pub contract TokenForwarding {
             emit ForwardedDeposit(amount: balance, from: self.owner?.address)
         }
 
+        pub fun check(): Bool {
+            return self.recipient.check<&{FungibleToken.Receiver}>()
+        }
+
+        pub fun safeBorrow(): &{FungibleToken.Receiver}? {
+            return self.recipient.borrow<&{FungibleToken.Receiver}>()
+        }
```
austinkline, line 57: Thinking through this.. Is there a scenario where this borrow panics? Isn't that what happens after secure cadence if the completely wrong type is underneath?

In that case, we will need to check first, and then return the borrow result afterwards

joshuahannan: that is a good point. I think you might be right. better to check beforehand

bluesign:
there is also good old cyclic link: (panics even on check)
```
acct.link<&AnyStruct>(/public/deniz, target: /public/bluesign)
acct.link<&AnyStruct>(/public/bluesign, target: /public/deniz)
```

bjartek:
is it possible to have this feature merged? We want to use Switchboard as royalty receiver in Doodles to handle royalties on both dapper and non dapper wallet, but I do not want to break it for @austinkline, and he needs this feature.

joshuahannan:
There are still a few unresolved comments so we can't merge it yet unfortunately

bluesign:
I am worried a bit on this `safeBorrow` , maybe we can add some checks in adding capability? otherwise if someone accidentally adds `&Vault` as receiver, borrower can take a reference to all vault.

joshuahannan:
The safe borrow method returns a restricted reference `&{FungibleToken.Receiver}`, so it can't return a vault reference, just a receiver reference

bluesign:
> The safe borrow method returns a restricted reference `&{FungibleToken.Receiver}`, so it can't return a vault reference, just a receiver reference

ah true, I though somehow it is just making `borrow()`

bluesign:
```
    pub fun safeBorrowByType(type: Type): &{FungibleToken.Receiver}? {
             if !self.checkReceiverByType(type: type) {
                 return nil
             }

             return self.receiverCapabilities[type]!.borrow()
         }
```

this one is problematic

joshuahannan:
@bluesign How is that problematic? It still only returns receiver references

bjartek:
any ETA on when this can be merged? It is needed.

joshuahannan:
We still need to figure out if @bluesign's comment is an issue before we can merge it. I don't believe it is a problem, but I just want to double check. In addition, there are still no tests in this PR, so I'm not comfortable merging it until there are some tests written.

bjartek:
```import FungibleToken from "../contracts/standard/FungibleToken.cdc"
import FungibleTokenSwitchboard from "../contracts/standard/FungibleTokenSwitchboard.cdc"
import FUSD from "../contracts/standard/FUSD.cdc"

transaction() {

	prepare(acct: AuthAccount) {
		// Get a Receiver reference for the Dapper account that will be the recipient of the forwarded DUC and FUT

		//FUSD
		let fusdReceiver = acct.getCapability<&{FungibleToken.Receiver}>(/public/fusdReceiver)
		if !fusdReceiver.check() {
			let fusd <- FUSD.createEmptyVault()
			acct.save(<- fusd, to: /storage/fusdVault)
			acct.link<&FUSD.Vault>( /public/fusdReceiver, target: /storage/fusdVault)
		}
		let switchboard <- FungibleTokenSwitchboard.createSwitchboard()
		switchboard.addNewVault(capability: fusdReceiver)


		if let typ = switchboard.safeBorrowByType(type: Type<@FUSD.Vault>()) {

			let vault = typ as! &FUSD.Vault
			//compile error

		}
		destroy switchboard
	}
}
```

bluesign:
ah yeah probably I remember wrong somehow, it works with `let fusdReceiver = acct.getCapability<&FUSD.Vault>(/public/fusdReceiver)` though.

austinkline:
> any ETA on when this can be merged? It is needed.

Yeah, the ball is in my court here I just haven't had the time to leave flowty work to finish this out with tests and some small alterations (doing a check before a borrow)

austinkline:
@bluesign hadn't thought about the downcasting issue, we could probably throw together a quick script to make sure what you're describing isn't possible, though. My understanding is that since borrow is has T set to `&{FungibleToken.Receiver}`, we should be good but definitely best to confirm

bluesign:
@austinkline yeah my example was little bad; 

I think situation is more like this: 

- I have a vault linked as `&FlowToken.Vault to `/private/denizVault`
- I can get capability to this: `getCapability<&FlowToken.Vault>(/private/denizVault)`
- `switchboard.addNewVault` will accept this as `&FlowToken.Vault` is also `&AnyResource{FT.Receiver}`
- then `switchboard.safeBorrowByType` will borrow this as `&FlowToken.Vault`
- it will return `&FlowToken.Vault` ( return type will be ok as it is also `&AnyResource{FT.Receiver}` ) 
- then who got this, can cast it to `&FlowToken.Vault`

I mean it is not a big deal, but also it is unexpected.

alilloig:
Wow @bluesign you have a twisted mind (in the best sense ofc ;) ), I love it!!  Honestly I had never thought of that, @turbolent could you please help us confirm that the specified type on a returning value is checking a conformance rather than restricting the type?

@austinkline would you mind if I made an issue and a new PR out of this adding some tests? I have done the rest of the tests for the switchboard so I should be fine. If you don't have any objection I'll do it on Monday so @bjartek can have it asap

austinkline:
> @austinkline would you mind if I made an issue and a new PR out of this adding some tests? I have done the rest of the tests for the switchboard so I should be fine. If you don't have any objection I'll do it on Monday so @bjartek can have it asap

Feel free!

joshuahannan:
I'm closing this because it is included in https://github.com/onflow/flow-ft/pull/125 now. That could use some more reviews if anyone has time!

Issue #113, opened 2023-01-19
Add Practical Token Setup and Transfer transactions that utilize switchboard


joshuahannan:
### Issue To Be Solved
Need more practical transactions for setting up the Fungible Token Switchboard in real user's accounts.
More context documented here: https://www.notion.so/dapperlabs/Royalties-Switchboard-393d11734bc64341a44131a7b3621053

### Suggest A Solution

Add new transactions:

* Update [setup_account](https://github.com/onflow/flow-ft/blob/master/transactions/setup_account.cdc) to set up a switchboard resource if it isn't already there and add the correct capability to it.
* Transactions to add a generic receiver capability or list capabilities to the switchboard provided a path or list of paths to receiver capabilities.
* Transaction to setup an account with USDC and add the USDC receiver to the switchboard
* Add a generic transfer transaction to check if the receiver provided is a switchboard and accepts the given token type, and if not, utilize the lost and found smart contract to store the temporary vault to be transferred. (No need to copy the whole lost and found contract to this repo to test this one, just do a manual test on the emulator to make sure it works)

cc @albeethekid

bjartek:
`import TokenForwarding from "../contracts/standard/TokenForwarding.cdc"
import FungibleToken from "../contracts/standard/FungibleToken.cdc"
import DapperUtilityCoin from "../contracts/standard/DapperUtilityCoin.cdc"
import FlowUtilityToken from "../contracts/standard/FlowUtilityToken.cdc"
import FungibleTokenSwitchboard from "../contracts/standard/FungibleTokenSwitchboard.cdc"
import FiatToken from "../contracts/standard/FiatToken.cdc"
import FUSD from "../contracts/standard/FUSD.cdc"
import FlowToken from "../contracts/standard/FlowToken.cdc"

/** 
 This is a transaction to set up an merchant account

 It has to be a blocto account since dapper will not allow us to run this account on a merchan account

 The only input parameter to this is your merchant account at dapper

 // 1. run this transaction in a lilico/blocto account your "credit account"
 // 2. use the path = /public/fungibleTokenSwitchboardPublic on this account as royalty receiver, it will be able to handle all the common FT
**/

transaction(dapperMerchantAccountAddress: Address) {

	prepare(acct: AuthAccount) {
		// Get a Receiver reference for the Dapper account that will be the recipient of the forwarded DUC and FUT
		let dapper = getAccount(dapperMerchantAccountAddress)

		//FUSD
[...]

PR #116, opened 2023-02-08
Include more transaction templates on Switchboard and a way for retrieving receivers along addresses 

alilloig:
Closes: #113 

## Description

______

For contributor use:

### DIFF
--- a/contracts/FungibleTokenSwitchboard.cdc
+++ b/contracts/FungibleTokenSwitchboard.cdc
@@ -36,6 +36,7 @@ pub contract FungibleTokenSwitchboard {
     /// 
     pub resource interface SwitchboardPublic {
         pub fun getVaultTypes(): [Type]
+        pub fun getVaultTypesWithAddress(): {Type: Address}
         pub fun deposit(from: @FungibleToken.Vault)
         pub fun safeDeposit(from: @FungibleToken.Vault): @FungibleToken.Vault?
     }
@@ -62,7 +63,7 @@ pub contract FungibleTokenSwitchboard {
             // Borrow a reference to the vault pointed to by the capability we 
             // want to store inside the switchboard
             let vaultRef = capability.borrow() 
-                ?? panic ("Cannot borrow reference to vault from capability")
+                        ?? panic ("Cannot borrow reference to vault from capability")
             // Check if there is a previous capability for this token, if not
             if (self.receiverCapabilities[vaultRef.getType()] == nil) {
                 // use the vault reference type as key for storing the 
@@ -71,8 +72,8 @@ pub contract FungibleTokenSwitchboard {
                 // emit the event that indicates that a new capability has been 
                 // added
                 emit VaultCapabilityAdded(type: vaultRef.getType(),
-                                    switchboardOwner: self.owner?.address, 
-                                    capabilityOwner: capability.address)
+                                               switchboardOwner: self.owner?.address, 
+                                                 capabilityOwner: capability.address)
             } else {
                 // If there was already a capability for that token, panic
                 panic("There is already a vault in the Switchboard for this token")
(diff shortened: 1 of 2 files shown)


bluesign:
I am a bit concerned about `getVaultTypesWithAddress`, what is the use case for that @alilloig  ?

contracts/FungibleTokenSwitchboard.cdc
```diff
@@ -126,18 +128,56 @@ pub contract FungibleTokenSwitchboard {
...
+        /// receivers such as a `@TokenForwarding.Forwarder`. It can also be
+        /// used to overwrite the types attached to certain capabilities without 
+        /// having to remove those capabilities first.
+        ///                    
+        /// @param paths: The paths where the public capabilities are stored.
+        /// @param types: The types of the fungible token to be deposited on each path.
+        /// @param address: The address of the owner of the capabilities.
+        /// 
+        pub fun addNewVaultsWrappersByPath(paths: [PublicPath], types: [Type], 
+                                                                  address: Address) {
```
bluesign, line 156: maybe `pub fun addNewVaultWrappersByPath` ?

bluesign: also why don't we send here [Type] and [Capability<&FungibleTokenReceiver>]

alilloig: yep that sounds better thx! 

Honestly I just thought it will be nice to have the same methods for capabilities stored under their own type and for wrappers, so I mixed `addNewVaultsByPath` and `addNewVaultWrapper`

alilloig:
> I mean Address there gives the impression that it is the final destination somehow. (instead of being capability address)

I totally agree with you. Do you think that a more specific name could help understand that those are not the final addresses but rather where the stored capabilities are pointing?

The use case is that I thought that if somebody is using this for royalties and have their different token vaults spread among different accounts (or even they have multiple instances of the same token vault in different accounts) could find useful a single method to check where its switchboard is redirecting the funds

alilloig:
I believe we should not include neither the USDC or the L&F contracts here. I haven't tested the transactions yet, I'm gonna do it on testnet as soon as @albeethekid can confirm he's happy with them

chandanworkacct (approved):

transactions/switchboard/safe_deposit_to_lnf.cdc
```diff
@@ -0,0 +1,80 @@
...
+
+        // Get a reference to the recipient's Switchboard Receiver
+        let switchboardRef = recipient.getCapability(FungibleTokenSwitchboard.PublicPath)
+            .borrow<&FungibleTokenSwitchboard.Switchboard{FungibleTokenSwitchboard.SwitchboardPublic}>()
+			?? panic("Could not borrow receiver reference to switchboard!")    
+        
+        // Attempt to deposit the USDC funds into the receiver's switchboard 
+        if let notDepositedVault <-switchboardRef.safeDeposit(from: <- self.sentVault.withdraw(amount: amount)) {
+            
+            // If a vault is returned, then they deposit didn't success, so we put
```
joshuahannan, line 54:
```suggestion
            // If a vault is returned, then their deposit didn't succeed, so we put
```

joshuahannan (approved):

joshuahannan:
I'm kind of agnostic about `getVaultTypesWithAddress()`. he is right that sometimes the address won't actually be the final destination, but usually it will be and I think that information could potentially be valuable to some people. I don't see it being a problem or something that could hurt users though, so it seems fine to include it.

joshuahannan:
Are we good to merge this once it has picked up the latest changes from master @alilloig ?

alilloig:
Let me run the transactions on testnet first and I'll merge it

joshuahannan:
@alilloig Thanks for the info!

PR #123, opened 2023-03-01
Implementation-of-flip-69

satyamakgec:
It addresses https://github.com/onflow/flips/blob/fungibleToken-standard-improvement/flips/20230206-fungible-token-vault-type-discovery.md

TODO:

- [ ] Update package.json
- [x] fix js-test framework issue


contracts/FungibleToken.cdc
```diff
@@ -109,6 +109,27 @@ pub contract interface FungibleToken {
...
+        /// Below is referenced from the FLIP #69 https://github.com/onflow/flips/blob/main/flips/20230206-fungible-token-vault-type-discovery.md
+        /// 
+        /// Returns the type of implementing resource i.e If `FlowToken.Vault` implements
+        /// this then it would return `[Type<@FlowToken.Vault>()]` and if any custom receiver
+        /// uses the default implementation then it would return empty array as its parent
+        /// resource doesn't conform with the `FungibleToken.Vault` resource.
+        ///
+        /// @return list of supported vault types by the implemented resource.
+        /// 
+        pub fun getSupportedVaultTypes() :[Type] {
```
joshuahannan, line 122:
```suggestion
        pub fun getSupportedVaultTypes(): [Type] {
```

contracts/FungibleToken.cdc
```diff
@@ -109,6 +109,27 @@ pub contract interface FungibleToken {
         /// @param from: The Vault resource containing the funds that will be deposited
         ///
         pub fun deposit(from: @Vault)
+
+        /// Below is referenced from the FLIP #69 https://github.com/onflow/flips/blob/main/flips/20230206-fungible-token-vault-type-discovery.md
+        /// 
+        /// Returns the type of implementing resource i.e If `FlowToken.Vault` implements
```
joshuahannan, line 115:
```suggestion
        /// Returns the Vault types that the the receiver is able to accept in its `deposit` method
        /// By default, it returns the concrete type of the implementing resource if it is a subtype of `FungibleToken.Vault`
        /// i.e If `FlowToken.Vault` implements
```

joshuahannan (commented):
I looked at the cadence code. Looks good! Can you also add implementations for the switchboard and token forwarders, and can you add a transfer transaction that accepts an address and public path as a parameter, gets the receiver reference from that path, then only deposits the tokens if the `receiver.getSupportedVaultTypes()` supports the type of the vault that the transaction is attempting to transfer? If it doesn't just deposit the vault back into the owner's vault

franklywatson (approved):

joshuahannan (approved):
I'm good with it. Lets share this on discord to get some community input. Make sure to say that we want to upgrade this on testnet and mainnet soon

contracts/FungibleToken.cdc
```diff
@@ -109,6 +109,30 @@ pub contract interface FungibleToken {
...
+        /// this then it would return `[Type<@FlowToken.Vault>()]` and if any custom receiver
+        /// uses the default implementation then it would return empty array as its parent
+        /// resource doesn't conform with the `FungibleToken.Vault` resource.
+        ///
+        /// Custom receiver implementations are expected to upgrade their contracts to add an implementation
+        /// that supports this method because it is very valuable for various applications to have.
+        ///
+        /// @return list of supported deposit vault types by the implementing resource.
+        /// 
+        pub fun getSupportedVaultTypes(): [Type] {
```
joshuahannan, line 125: Thinking about this more, I would like the return type of this to be `{Type: Bool}`. You can still iterate through it with `return.keys`, and it makes checking a specific type much easier. Do you think you'll be able to make that change in time for the upgrade next week?

joshuahannan:
@satyamakgec You'll need to pull the latest from master before you push your new changes since it is out of date

satyamakgec:
@joshuahannan latest pull have taken and the return type also changed

satyamakgec:
May need to fix the Go tests to make the CI happy

joshuahannan (commented):
I'll fix the Go tests right now

joshuahannan (approved):

PR #127, opened 2023-03-16
Make return type to dictionary for discoverability of vault types

satyamakgec:
Extension to #123


joshuahannan (approved):

joshuahannan (approved):

bluesign:
> Dictionaries are more usable than arrays because you can index into them without having to iterate through the array to find an id, and you can still get an array from them anyway if you want by using `dictionary.keys`

But in this context it is only used for if item exists, which is `Array.Contains` in Cadence. 

`getSupportedVaultTypes().Contains(t)` 
vs
`getSupportedVaultTypes()[t]!=nil`

or even 

```cadence 
if let exists = getSupportedVaultTypes(t){
   if t == true{
  }
}
```

somehow I feel it doesn't make sense here.

bluesign:
> `array.contains` still has to iterate through the array under the hood to find the value, right? That is a O(n) operation, but with a dictionary, it is an O(1) operation I believe.

yeah but it matters only for big N; here we have something like N<10 maybe on worst case ( on most cases N=1 ) ( even iterating can be cheaper for some N < x )

joshuahannan:
but it would matter for something like the NFT standard, and I want to make the function signatures as similar as possible for these kinds of methods so using them can be similar. We're going to add something like this for NFTs also

PR #137, opened 2023-06-27
Switch testing to cadence for `ExampleToken` contract

m-Peter:
Work towards: https://github.com/onflow/developer-grants/issues/148

## Description

Uses the Cadence testing framework to test the `ExampleToken` smart contract.

______

For contributor use:

- [x] Update the version in package.json when there is a change in the smart contracts


Makefile
```diff
@@ -1,10 +1,7 @@
 .PHONY: test
 test:
-	$(MAKE) generate -C lib/go
-	$(MAKE) test -C lib/go
```
joshuahannan, line 4: I don't want to get rid of the Go packages. Other repos depend on them, so we need to leave these in here in addition to the Cadence tests, same with ci

tests/scripts/get_token_metadata.cdc
```diff
@@ -0,0 +1,16 @@
+import ExampleToken from "ExampleToken"
+import FungibleTokenMetadataViews from "FungibleTokenMetadataViews"
+import MetadataViews from "MetadataViews"
+
```
joshuahannan, line 4: Can you please add a short description at the top of each transaction in a comment? Thanks!

tests/scripts/get_token_metadata.cdc
joshuahannan, line 1: Is is possible to use the scripts and transactions that [we already have written](https://github.com/onflow/flow-ft/blob/master/transactions/scripts/metadata/get_token_metadata.cdc) instead of having different copies of them for tests?

tests/scripts/get_vault_data.cdc
```diff
@@ -0,0 +1,23 @@
...
+    let vaultData = FungibleTokenMetadataViews.getFTVaultData(vaultRef)
+        ?? panic("Token does not implement FTVaultData view")
+
+    let vault <- vaultData.createEmptyVault()
+
+    let vaultIsEmpty = vault.balance == 0.0
+
+    destroy vault
+
+    return vaultIsEmpty
```
joshuahannan, line 22: Not sure what you're doing here. You get the `vaultData` view, then don't use it and instead just return a boolean about whether a newly created empty vault is empty. Should this be different?

tests/test_example_token.cdc
```diff
@@ -0,0 +1,157 @@
...
+
+    let result = blockchain.executeTransaction(tx)
+
+    Test.expect(result, Test.beSucceeded())
+
+    var typ = CompositeType("A.01cf0e2f2f715450.ExampleToken.TokensDeposited")!
+    Test.assertEqual(2, blockchain.eventsOfType(typ).length)
+
+    typ = CompositeType("A.01cf0e2f2f715450.ExampleToken.TokensWithdrawn")!
+    Test.assertEqual(1, blockchain.eventsOfType(typ).length)
```
joshuahannan, line 93: Can you add scripts here to check here if the recipient's balance increased by 50 and the sender's balance decreased by 50? Thank you!

tests/test_example_token.cdc
```diff
@@ -0,0 +1,157 @@
...
+
+    Test.expect(result, Test.beSucceeded())
+}
+
+pub fun testGetTokenMetadata() {
+    let code = Test.readFile("./scripts/get_token_metadata.cdc")
+    let result = blockchain.executeScript(code, [recipient.address])
+
+    Test.expect(result, Test.beSucceeded())
+}
```
joshuahannan, line 150: It would be great to verify the return values here too to make sure they are correct

tests/test_example_token.cdc
joshuahannan, line 1: The test setup looks good, but it would also be great to have some negative test cases verified too, such as if an account tries to send more tokens than they have in their vault, etc

joshuahannan (commented):
Thank you so much for making these changes! This is way cleaner. I just had a few comments but everything looks pretty good so far. 😃

Makefile
m-Peter: Makes sense, I will bring them back, and fix the tests as well, with the updated imports :pray:

tests/scripts/get_token_metadata.cdc
m-Peter: Added in https://github.com/onflow/flow-ft/pull/137/commits/d7a27dc816bf01cad2082ca4aeaec2b781e9d969

tests/scripts/get_token_metadata.cdc
m-Peter: Unfortunately, this is probably one the last limitations of the Cadence testing framework, it is not possible yet to return on-chain types to the test files. However, I have updated the copies to actually test the return values. It's one of the current workarounds. For example: https://github.com/onflow/flow-ft/pull/137/commits/d7a27dc816bf01cad2082ca4aeaec2b781e9d969#diff-3548e7dc61554c9b9f3f368ac783f2101b3bc8c61171c5ef974a51561b3ab3daR22-R35
Let me know what you think :pray:

tests/scripts/get_vault_data.cdc
m-Peter: I have reworked this test in https://github.com/onflow/flow-ft/pull/137/commits/d7a27dc816bf01cad2082ca4aeaec2b781e9d969#diff-3548e7dc61554c9b9f3f368ac783f2101b3bc8c61171c5ef974a51561b3ab3daR22-R35. Basically, I exercise the `vaultData.createEmptyVault()` closure, in order to increase the coverage, and make sure that the closure actually works as expected.

tests/test_example_token.cdc
m-Peter: Nice catch :100:
I have added more assertions in this test case: https://github.com/onflow/flow-ft/pull/137/commits/d7a27dc816bf01cad2082ca4aeaec2b781e9d969#diff-39b918cf828753efa2720618fdc579fe01b8bec8251d8c42c039a7e6ae9ef7cdR137-R160

tests/test_example_token.cdc
m-Peter: Makes sense :+1: 
I have added assertions in the test scripts: https://github.com/onflow/flow-ft/pull/137/commits/d7a27dc816bf01cad2082ca4aeaec2b781e9d969#diff-5c07d50f7e7ec4083155d1121b5006621428e4ebfe2a193475ec372126a6c631R21-R26

tests/test_example_token.cdc
m-Peter: Nice catch :100: There are indeed more test cases to be covered. I opened this PR in order to pave the way :pray: 
Added in https://github.com/onflow/flow-ft/pull/137/commits/d7a27dc816bf01cad2082ca4aeaec2b781e9d969#diff-39b918cf828753efa2720618fdc579fe01b8bec8251d8c42c039a7e6ae9ef7cdR162-R177. I have also added many more assertions for most of the test cases.
Please have another look :pray:

Makefile
m-Peter: Updated in https://github.com/onflow/flow-ft/pull/137/commits/4adf07d3d7c64e59116fd6ffc49ed9fb064bb806

joshuahannan (approved):
Thanks for addressing my comments! The changes look really good. Did all the tests pass when you ran `make test`? I need to see why CI isn't running here for you

m-Peter:
@joshuahannan Both `make test` & `make ci` pass locally, and in the GitHub workflow. However, this is a fork, probably that's why it is not running here, but rather in our fork: https://github.com/Build-Squad/flow-ft/actions/runs/5401591136

PR #139, opened 2023-07-11
Adds total supply view and updates cadence tests

joshuahannan:
## Description

* Adds a view for the total supply of a token
* Updates cadence, emulator, and sdk dependencies. All the changes to the tests and templates files don't need to be reviewed because they are just updates to dependencies.
* Removes JS tests and add Cadence testing framework tests for the switchboard

______

For contributor use:


joshuahannan:
@sisyphusSmiling @satyamakgec I'm having some issues with the JS test setup in this PR when I run the tests and when CI runs. Can one of you take a look? Thank you!

sisyphusSmiling:
cc: @jribbink @franklywatson could use some js-friendly eyes

Not sure why the js tests are failing here. I'm getting some odd behavior - `beforeEach()` throws undefined and every other test fails with connection refused:

```
HTTP Request Error:
          HTTP Request Error: An error occurred when interacting with the Access API.
          error=request to http://localhost:57873/v1/network/parameters failed, reason: connect ECONNREFUSED 127.0.0.1:57873
          hostname=http://localhost:57873
          path=/v1/network/parameters
          method=GET
```

I was able to get `switchboard.test.js` passing at some point and now they're all failing. For reference, here's my `npm version `output:

```
{
  test: '1.0.0',
  npm: '9.8.0',
  node: '16.16.0',
  v8: '9.4.146.24-node.21',
  uv: '1.43.0',
  zlib: '1.2.11',
  brotli: '1.0.9',
  ares: '1.18.1',
  modules: '93',
  nghttp2: '1.47.0',
  napi: '8',
  llhttp: '6.0.7',
  openssl: '1.1.1q+quic',
  cldr: '40.0',
  icu: '70.1',
  tz: '2021a3',
  unicode: '14.0',
  ngtcp2: '0.1.0-DEV',
  nghttp3: '0.1.0-DEV'
}
```

This is after updating to latest stable npm release from `8.11.0`

joshuahannan:
@sisyphusSmiling and @satyamakgec I removed the JS tests and added some cadence tests for the switchboard, which was the only extra thing that wasn't already handled by the Go/Cadence tests, so this is ready for another review

transactions/scripts/metadata/get_vault_supply_view.cdc
```diff
@@ -0,0 +1,20 @@
...
+
+pub fun main(address: Address): UFix64 {
+    let account = getAccount(address)
+
+    let vaultRef = account
+        .getCapability(ExampleToken.VaultPublicPath)
+        .borrow<&{MetadataViews.Resolver}>()
+        ?? panic("Could not borrow a reference to the vault resolver")
+
+    let ftSupply = vaultRef.resolveView(Type<FungibleTokenMetadataViews.TotalSupply>())!
```
sisyphusSmiling, line 15: np: Maybe not in scope for this PR, but we could generalize this script if `ExampleToken` implemented `ViewResolver`. I think we should encourage NFT & FT contracts to implement that interface anyway.

sisyphusSmiling (approved):
LGTM. Glad to see we're getting more Cadence tests in the mix!

transactions/scripts/metadata/get_vault_supply_view.cdc
joshuahannan: Why would `ExampleToken` need to implement `ViewResolver` to make this generalizable? Isn't it already generalizable if you just provide the path as an argument instead of hardcoding? Agreed about making it implement `ViewResolver` though.

sisyphusSmiling: Yeah, totally true. Could just add public path. I've just come across a couple of instances with HC scripts/txns where `ViewResolver` would make things easier, so I've been more inclined to spot opportunities to encourage the pattern. You're right though, either one works

PR #145, opened 2023-09-06
Remove restricted types from link

bjartek:
when deploying this against stable cadence you cannot have restricted types.


joshuahannan (approved):
Thanks! I haven't fully gotten to testing this ExampleToken, so there are probably some other issues too, but I'll be working on it more soon. I really appreciate the help. 😃

PR #149, opened 2023-10-18
Update repo contents for Cadence 1.0

sisyphusSmiling:
Related: #96 

## Description

Updates this repo's contracts, transactions & scripts for Cadence 1.0 with passing go and cadence tests. README docs will follow in a successive PR.
______

For contributor use:


sisyphusSmiling:
@m-Peter do you know why the error is occurring in the `example_token_tests.cdc`

```sh
error: internal error: could not find account with address 0000000000000005
```

I believe this is occurring when attempting to construct a `CompositeType` using the deployment `admin` account defined in the setup configuration since all test cases attempting to do so fail.

e.g.
```cadence
access(all) fun testTokensInitializedEventEmitted() {
    let typ = CompositeType(buildTypeIdentifier(admin, "ExampleToken", "TokensInitialized"))
        ?? panic("Problem constructing CompositeType")
    Test.assertEqual(1, blockchain.eventsOfType(typ).length)
}

// Test case calls through to this helper function which I think causes an error at acct.address
access(all) fun buildTypeIdentifier(_ acct: Test.TestAccount, _ contractName: String, _ suffix: String): String {
    let addrString = acct.address.toString()
    return "A.".concat(addrString.slice(from: 2, upTo: addrString.length)).concat(".").concat(contractName).concat(".").concat(suffix)
}
```

Any thoughts on this one?

m-Peter:
Hmm, I recall trying the 1st preview release of `flow-cli`, with Stable cadence, but I did not have any problems.
Probably the next releases included some of the merged work in `cadence-tools/test` repository, but I do not at which state.
If you can wait some time, I have started the migration here: https://github.com/onflow/flow-ft/pull/150

contracts/ExampleToken-v2.cdc
```diff
@@ -18,6 +18,7 @@ access(all) contract ExampleToken: ViewResolver {
     /// User Paths
     access(all) let VaultStoragePath: StoragePath
     access(all) let VaultPublicPath: PublicPath
+    access(all) let ReceiverPublicPath: PublicPath
```
joshuahannan, line 21: Why add a receiver public path? We already have the vault public path. With entitlements, there is no difference any more

contracts/ExampleToken-v2.cdc
```diff
@@ -215,6 +231,9 @@ access(all) contract ExampleToken: ViewResolver {
         /// and returns them to the calling context.
         ///
         access(all) fun mintTokens(amount: UFix64): @ExampleToken.Vault {
+            pre {
+                amount > 0.0: "Amount minted must be greater than zero"
+            }
```
joshuahannan, line 236: Let's not include this because it is kind of unnecessary

contracts/utility/MetadataViews.cdc
joshuahannan, line 1: You just copied this from the NFT repo, right? You didn't make any changes to it that are different than what is there?

joshuahannan (commented):
Just leaving a few preliminary comments

contracts/utility/MetadataViews.cdc
sisyphusSmiling: Yes, copied from https://github.com/onflow/flow-nft/tree/standard-v2

contracts/ExampleToken-v2.cdc
sisyphusSmiling: In the context of this single contract, I agree - may as well just use a single storage & public path.

But when incorporating `TokenForwarding`, I ran into an issue wiring up the' `Receiver` implementing `Forwarder` resource at the `ExampleToken.VaultPublicPath`. `Forwarder` doesn't conform to the `Vault` interface so it can only be linked as a `Receiver` and doing so at the `.VaultPublicPath` breaks expectations for getters trying to retrieve `&{FungibleToken.Vault}`.

I thought creating another canonical path dedicated to the `Receiver` functionality would allow for more granularity in cases like these.

Does that make sense?

joshuahannan: that is a good point. Might need to consider adding that to the standard itself and adding a comment explaining why we still need both paths. I guess it is kind of similar to how we have the receiver public path and the balance public path as separate paths in the v1 standard. Can you make that change in the standard?

sisyphusSmiling: I don't think we can add those paths as contract fields in the v2 standard. FungibleToken will no longer be a contract interface, so those values would need to be initialized and lead to collisions between vaults, wouldn't they?

joshuahannan: thank you!

sisyphusSmiling:
Thanks for pointing out #150 @m-Peter. Probably best for this PR to wait for that to be approved & merged so this branch can pick up the test changes.

contracts/ExampleToken-v2.cdc
```diff
@@ -232,6 +247,16 @@ access(all) contract ExampleToken: ViewResolver {
         return <- create Vault(balance: 0.0)
     }
 
+    /// Function that destroys a Vault instance, effectively burning the tokens.
+    ///
+    /// @param from: The Vault resource containing the tokens to burn
+    ///
+    // TODO: Revisit if removal of custom destructors passes
```
joshuahannan, line 254:
```suggestion
    // TODO: Revisit if removal of custom destructors passes
    // Will need to add an update total supply
```

tests/metadata_views_tests.cdc
```diff
@@ -0,0 +1,90 @@
...
+
+    Test.assertEqual(true, result)
+}
+
+
+/* Transaction Helpers */
+
+access(all) fun setupExampleToken(_ acct: Test.TestAccount) {
+    txExecutor("setup_account.cdc", [acct], [], nil, nil)
+}
```
joshuahannan, line 39: I see this and other helpers in multiple test files. Would it make sense to put them in a single file so we don't have duplicate code?

joshuahannan (approved):
Really great work on these! Thank you so much for working on all of this. I just had a couple comments. 

Also, I merged https://github.com/onflow/flow-ft/pull/150 to `master`, so you might want to check that your tests are still compatible with the changes there. Same with the NFT repo tests as well

PR #155, opened 2024-03-26
Add `nil` check for Burner

joshuahannan:
## Description

Add check to see if `AnyResource` is nil before burning it with the Burner

______

For contributor use:

- [x] Targeted PR against `v2-standard` branch


coverage.json
sisyphusSmiling, line 1: I think this file should be included in `.gitignore`. If so, np to include in separate PR

sisyphusSmiling (approved):
Looks good. Only note is unrelated - looks like the repo isn't configured for codecov which is an option now that we have Cadence tests. I can get that setup if you like, just lmk

coverage.json
joshuahannan: it is in there already, just hadn't removed it yet. I removed it

joshuahannan:
I think this is fine for now. If we see that double optionals are causing problems in the future, then we can revisit

PR #166, opened 2024-07-09
Add assertions to generic transfer and update docs

joshuahannan:
Closes: #162 

## Description

* Adds assertions to `generic_transfer_with_address` to make sure the vault is the correct type
* Added a test with `MaliciousToken` to test the transaction. `MaliciousToken` is the same as `ExampleToken`, but returns the `ExampleToken` paths in the `FTVaultData` view instead of its own paths.
* Adds FLIX to the main generic transactions
* Updates comments in `FungibleToken`
* Updates tests to use the pre-deployed versions of the contracts that do not come from this repo instead of deploying new ones

I can't think of any other assertions to put in the transactions, but I am open to suggestions

______

For contributor use:


transactions/generic_transfer_with_address.cdc
```diff
@@ -31,6 +44,23 @@ transaction(amount: UFix64, to: Address, contractAddress: Address, contractName:
 			?? panic("Could not borrow reference to the owner's Vault!")
 
         self.tempVault <- vaultRef.withdraw(amount: amount)
+
+        // Get the string representation of the address without the 0x
+        var addressString = contractAddress.toString()
+        if addressString.length == 18 {
+            addressString = addressString.slice(from: 2, upTo: 18)
+        }
+        let typeString: String = "A.".concat(addressString).concat(".").concat(contractName).concat(".Vault")
```
sisyphusSmiling, line 53: This covers all existing `Vault` implementations but wouldn't cover multi-vault FT contracts. We should probably note that in the docs and add a `generic_transfer_with_identifier` as was suggested for bridge transactions

sisyphusSmiling (approved):
Sorry for the delay on this review. Left one non-blocking comment

transactions/generic_transfer_with_address.cdc
joshuahannan: good point! I created an issue here:https://github.com/onflow/flow-ft/issues/168

PR #178, opened 2025-04-23
Moves burnCallback default implementation to the `Balance` interface

joshuahannan:
## Description

A recent update to Cadence made it so that pre-conditions in a default implementation do not execute if the implementation is overridden. This moves the default implementation for `burnCallback` to the `Balance` interface so that the pre conditions will always execute.

Need to be sure that this upgrade won't have any negative affects:

* Those who were relying on the default implementation will still have it via the `Balance` interface
* If there were any resources that implemented the `Balance` interface, which there probably weren't, then the addition of `burnCallback` and its default implementation will not be a problem because it is `access(contract)`, so only the `Burner` and `FungibleToken` contracts can call it.

______

For contributor use:


bluesign:
> A recent update to Cadence made it so that pre-conditions in a default implementation do not execute if the implementation is overridden.

this sounds like a bug in Cadence imo

turbolent:
The changes look good to me 👍 

To answer the concerns regarding the behaviour:

An interface function with only conditions, and no statements, means the conditions are inherited in the implementation function. 

If an interface functions has statements, then it is considered a default function. 

If an implementation of the interface (type conforms to the interface) defines a function, then it is used *instead* of the default function, it *overrides* the default function. 

This intended behaviour was unfortunately not correctly implemented in Cadence, and conditions of default functions were included when executing the function in the implementation of the interface.

This bug was recently fixed (https://github.com/onflow/cadence/pull/3775, in particular https://github.com/onflow/cadence/pull/3775/commits/6f24819faa231ae0170d0a859364079477c3cb20). 

Existing code might have relied on this wrong behaviour, for example the Fungible Token.

Unfortunately, it is not clearly documented what it means to define an interface function that has conditions and statements, neither in the documentation nor in FLIPs, which leads to users assuming certain behaviour – either that conditions are inherited; or they are not, and only apply to the default function.

Users might want to both define conditions that should be executed for every implementation, and at the same time also provide a default implementation.

One can argue for either behaviour (conditions being inherited or not), and neither is more "correct" than the other. For now we rectified the intended behaviour in the implementation. We  will also look into how we can make this intent (providing a default function and also enforcing conditions) less error-prone / a footgun, e.g. by improving documentation, tooling (linting), and considering language improvements.

For now, this intent can be best implemented in a Cadence program by having two interfaces, i.e. splitting the function interface function with conditions and statements into two parts: One interface has the conditions that should be inherited in all implementations, and another interface that provides the default function.

bluesign:
thanks @turbolent

yeah this is tricky; I think current implementation is the one that can make sense ( considering post conditions )

contracts/FungibleToken.cdc
```diff
@@ -77,6 +77,13 @@ access(all) contract interface FungibleToken: ViewResolver {
     ///
     access(all) resource interface Balance {
```
sisyphusSmiling, line 78: It feels like this should inherit the `Burner.Burnable` interface. Not a big deal in the context of `Vault` implementations, but I'm wondering about other implementations that want a balance (e.g. a fractional NFT or Game resource).

For instance, let's say I have something like an in-game resource that has a balance.


```cadence
access(all) resource NFT : FungibleToken.Balance, ... {
  access(all) var balance: UFix64
  ...
}
```

The NFT will inherit `Balance` and therefore the default `burnerCallback` method. But when the NFT is burned, `Burner.burn` won't actually execute the `burnCallback` due to [this line in `Burner`](https://github.com/onflow/flow-ft/blob/master/contracts/utility/Burner.cdc#L30)

```cadence
if let s <- r as? @{Burnable} 
```

So in the end, the NFT will just be destroyed and the `Balance.burnCallback()` won't be called. I think if we're going to include functionality that's explicitly related to the `Burnable` interface, it should probably be inherited here. But I'm not 100% convicted and I'm curious to hear other thoughts.

contracts/FungibleToken.cdc
```diff
@@ -77,6 +77,13 @@ access(all) contract interface FungibleToken: ViewResolver {
...
         access(all) var balance: UFix64
+
+        // This default implementation needs to be in a separate interface
+        // from the one in `Vault` so that the conditions get enforced
+        // in the correct one
+        access(contract) fun burnCallback() {
```
sisyphusSmiling, line 84:
```suggestion
        /// This default implementation needs to be in a separate interface
        /// from the one in `Vault` so that the conditions get enforced
        /// in the correct one.
        /// Note: Anyone implementing `Balance` without also implementing
        /// `Vault` should also inherit `Burner.Burnable` for this method to
        /// be executed when the resource is passed to `Burner.burn()`
        access(contract) fun burnCallback() {
```

sisyphusSmiling (approved):

contracts/FungibleToken.cdc
joshuahannan: I added the `Burnable` inheritance to `Balance`, so this won't be an issue

